PR #5, opened 2019-04-28
Tests

kvikshaug:
- Fix firebase warnings
- Add mocked store ([docs](https://vue-test-utils.vuejs.org/guides/#testing-vuex-in-components)) - fixes `TypeError: Cannot read property 'state' of undefined`
- The only remaining issue is that `localStorage` needs to be mocked. The test is being skipped for now until that is added.

Note: Firebase still yields the warning below, I don't really see a way to suppress it:

```
Warning: This is a browser-targeted Firebase bundle but it appears it is being
run in a Node environment.  If running in a Node environment, make sure you
are using the bundle specified by the "main" field in package.json.

If you are using Webpack, you can specify "main" as the first item in
"resolve.mainFields":
https://webpack.js.org/configuration/resolve/#resolvemainfields

If using Rollup, use the rollup-plugin-node-resolve plugin and set "module"
to false and "main" to true:
https://github.com/rollup/rollup-plugin-node-resolve
```

### DIFF
--- a/src/components/LoginDialog.vue
+++ b/src/components/LoginDialog.vue
@@ -145,7 +145,7 @@ import axios from "axios";
 import { AxiosResponse } from "axios";
 import settings from "@/settings";
 import { JWT } from "@/store/session";
-import * as firebase from "firebase/app";
+import firebase from "firebase/app";
 import "firebase/auth";
 
 export default Vue.extend({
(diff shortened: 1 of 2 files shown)


kvikshaug:
> One thing I noticed is that in the docs you linked, https://vue-test-utils.vuejs.org/guides/#mocking-actions they use special `jest.fn()` functions to mock certain behavior. Do we need a mocha/chai equivalent rather than the empty pure functions that you're using?

That will be useful when we want to explicitly test that actions are being called. That's not the case yet though, the current test really only mounts the component.

PR #7, opened 2019-04-29
Designs

annasirunian:
1. Designs table (todo: restyle columns in expanded rows)
2. Shared Deletion Component

### DIFF
--- a/src/components/Designs.vue
+++ b/src/components/Designs.vue
@@ -14,6 +14,7 @@
           :headers="headers"
           :items="designs"
           :expand="expand"
+          :pagination.sync="pagination"
           select-all="primary"
           class="elevation-8"
         >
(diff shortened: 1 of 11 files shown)


src/components/shared/NewProject.vue
```diff
@@ -80,6 +80,7 @@ export default Vue.extend({
   methods: {
     createProject() {
       this.$store.commit("toggleDialog", "loader");
+      const a = this.projectName.value;
```
kvikshaug: Leftover from debugging?

src/components/shared/DeletionDialog.vue
```diff
@@ -0,0 +1,110 @@
...
+      axios
+        .all(
+          this.items.map(item =>
+            axios.delete(
+              `${settings.apis[storage]}/${this.itemsType}/${item.id}`
+            )
+          )
+        )
+        .then(
+          axios.spread(aaa => {
```
kvikshaug: This should probably be called `acct` to match the parameter, or just omitted if not in use

annasirunian: yes, thanks, will fix that

src/components/Designs.vue
```diff
@@ -0,0 +1,339 @@
...
+        isReactionKnockin && method === "Pathway"
+          ? JSON.parse(reaction).bigg_id
+          : reaction;
+      if (reactionId.startsWith("MNX")) {
+        return `https://www.metanetx.org/equa_info/${reactionId}`;
+      }
+      return `http://bigg.ucsd.edu/search?query=${reactionId}`;
+    },
+    geneLink(geneId) {
+      return `http://bigg.ucsd.edu/search?query=${geneId}`;
```
kvikshaug: Not a big deal but I don't really feel this needs to be a method, just put the URL directly in the `href`.

src/components/shared/DeletionDialog.vue
```diff
@@ -0,0 +1,110 @@
...
+      const storage = {
+        designs: "designStorage",
+        maps: "maps",
+        models: "modelsStorage",
+        projects: "iam"
+      }[this.itemsType];
+
+      const ids = this.items.map(item => item.id);
+      axios
+        .all(
```
kvikshaug: Nice use of concurrent requests :+1:

src/components/Designs.vue
```diff
@@ -0,0 +1,339 @@
...
+  name: "Designs",
+  data: () => ({
+    selected: [],
+    expand: true,
+    showAllReactionKnockins: false,
+    showAllReactionKnockouts: false,
+    showAllGeneKnockouts: false,
+    isDeletionDialogVisible: false,
+    isLoading: false,
+    headers: [
```
Midnighter: Your column widths only add up to 95%, could that be a reason for layout problems when you expand?

annasirunian: Oh, thank you, I should change it to 100%, initially I left 5% for checkbox area, but I couldn't specify width in the same way, expanded rows still need restyling though

PR #8, opened 2019-05-01
Maps and models

ChristianLieven:
To lower the threshold of reviewing this PR I'm gonna walk you through the general pattern on the example of Models CRUD.

1. Visuals:
The style is following the original CRUD views closely with only minor changes which I'll point out.
Central table with two columns:
Screenshot 2019-05-02 at 16 41 22

When the user is logged in and tried to edit public data the buttons-icons are disabled and a tooltip explains whats going on:
Screenshot 2019-05-02 at 16 41 47

When the user is logged out the FAB is disabled and a tooltip explains whats happening. Note here that I had to add special styling to the FAB ` v-bind:style="styleObject"` with ` styleObject: {
      "pointer-events": "auto"
    } ` because in `vuetify` removes all pointer events on disabled objects which prevents the tooltip from triggering:
Screenshot 2019-05-02 04 42 06

When adding or editing a model all form items are validated against a very simple rule `:rules="[rules.required]"` which is defined as `rules: {
      required: value => !!value || "Required."
    }`. I'm thinking of moving all the rule variables into the store or some config as they are consistently needed for all input fields and are mostly static. We might want to have more elaborate rules in the future (specific email validation or length limitations). Similarily, I've created a small object to store different tooltip messages as they are also recurrent and ideally I'd want to have to change them in one place only (i.e. `$store.state.commonTooltipMessages.publicData`).
Screenshot 2019-05-02 at 16 43 11

Visual change from the previous platform, instead of using a hidden `<p></p>`-tag to display that a model file is not valid, I'm using the assistive elements of the text field (https://material.io/design/components/text-fields.html#anatomy). I can of course change that back if so required.
Screenshot 2019-05-02 at 17 56 08

As for the input field, I'm using an input field which was introduced by Dohomi in https://github.com/vuetifyjs/vuetify/issues/238 as a replacement for a dedicated vuetify import component that is `coming soon(TM)`.

Another visual change, is the fact that the dialogs will layer. This was not an active choice but is the default behaviour of `vuetify`. I could try to invest time in trying to change it to match the old behaviour, although I do appreciate the visual continuity this creates i.e. a user knows exactly where they came from and what they'll go back to.
Screenshot 2019-05-02 at 18 18 56

2. Code:
All New`*` components emit two events:
- `this.$emit("input", value)` as part of `v-model` to communicate the `isVisible` boolean to the parent component triggering that the dialog should be closed.
- `this.$emit("returnObject", response.data);` when `*` has successfully been created. Components that need the `*` object right after it was created can listen to this.
All of these shared components also take the property `value` which completes the requirement to use `v-model`.


src/store/models.ts
```diff
@@ -36,7 +45,11 @@ export default {
   },
   getters: {
     getModelById: state => (id: number) => {
-      return state.models.find(model => model.id === id);
+      let model = state.models.find(model => model.id === id);
+      if (model === undefined) {
+        model = { name: "???" };
+      }
+      return model;
```
kvikshaug: I know I added this here - I'm thinking though that the handling of unrecognized models should go in the corresponding components, not the general getter, so that each component may decide how best to deal with it.

src/components/shared/NewProject.vue
```diff
@@ -68,26 +67,28 @@ import settings from "@/settings";
...
+        .post(`${settings.apis.iam}/projects`, payload)
         .then((response: AxiosResponse) => {
-          const projectItem = Object.assign(
-            { name: this.projectName.value },
-            response.data
-          );
-          this.$store.commit("projects/addProject", projectItem);
+          const commitPayload = Object.assign(payload, response.data);
+          console.log("the Project name:");
+          console.log(this.projectName);
```
kvikshaug: Remove console.log

src/components/shared/NewOrganism.vue
```diff
@@ -92,35 +94,27 @@ import settings from "@/settings";
...
   }),
   methods: {
     createOrganism() {
       this.$store.commit("toggleDialog", "loader");
-      const payload = {
-        name: this.organismName.value,
-        project_id: this.projectItemValidation.projectItem.id
-      };
+      const payload = { name: this.organismName, project_id: this.project.id };
+      console.log("The current organismItem:", payload);
```
kvikshaug: Remove

Midnighter:
Is there still a type error? I had found the reason for the last one.

ChristianLieven:
> Is there still a type error? I had found the reason for the last one.

Yeah no worries, I didn't get around to fixing it! Will take care before I merge this!

src/components/shared/NewProject.vue
```diff
@@ -52,10 +52,9 @@
     <v-snackbar
       color="success"
       v-model="isProjectCreationSuccess"
-      bottom
       :timeout="3000"
     >
-      {{ projectName.value }} successfully created.
+      {{ projectName }} successfully created.
```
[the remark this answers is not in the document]
ChristianLieven: This is an odd bug I presume with the snackbar component. I'll investigate this more in the polishing phase if thats OK?

ChristianLieven:
> * Looks like the model was empty when it was created:

Hmm it doesn't seem to be added to the table straight away. If you refresh it'll show up nicely. I'll have to think about a way to do that.

ChristianLieven:
> * Clicking edit/delete for public data still shows the modal even though they're disabled

Yikes, fixing that.

ChristianLieven:
> * When a project is created (as part of uploading model or map), we need to refresh the user's token, so that they get access to create the new model/map in that project

How do I go about that? As part of your work on this is there a function I can call to refresh this?

ChristianLieven:
> > * Clicking edit/delete for public data still shows the modal even though they're disabled
> 
> Yikes, fixing that.

Done!

kvikshaug:
> How do I go about that? As part of your work on this is there a function I can call to refresh this?

Almost - with [this](https://github.com/DD-DeCaF/caffeine-vue/pull/20) small modification, you can now dispatch the `session/refreshToken` action which will refresh the token.

You could _hope_ that the token refreshes by the time the user finishes the form - but if you wanted to wait for it, you can access the request promise on `state.session.refreshRequest`.

PR #10, opened 2019-05-01
chore: move store modules to sub directory

Midnighter:
I think it makes the organizational layout clearer.


Midnighter:
Let me check. I don't think the automatic namespace is tied to the directory structure.

Midnighter:
We don't have tests yet but log in and log out work, as well as fetching the resources from the backend. So I'm pretty sure everything is fine.

ChristianLieven:
Then it's good to go!

PR #11, opened 2019-05-01
Jobs

annasirunian:
Jobs list component

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "material-design-icons-iconfont": "^3.0.3",
     "roboto-fontface": "*",
     "vue": "^2.6.6",
+    "vue-moment": "^4.0.0",
     "vue-router": "^3.0.6",
     "vuetify": "^1.5.13",
     "vuex": "^3.0.1"
--- a/src/components/JobDetails.vue
+++ b/src/components/JobDetails.vue
@@ -1,0 +1,11 @@
+<template
+  ><p></p
+></template>
+
+<script lang="ts">
+import Vue from "vue";
+
+export default Vue.extend({
+  name: "JobDetails"
+});
+</script>
--- a/src/components/Jobs.vue
+++ b/src/components/Jobs.vue
@@ -1,7 +1,7 @@
 <template>
   <v-container>
     <v-layout justify-center>
-      <v-flex md8>
+      <v-flex md10>
         <h1 class="mb-2">Jobs</h1>
         <v-data-table
           :headers="headers"
@@ -18,6 +18,13 @@
             <td>{{ model(props.item.model_id).name }}</td>
             <td>{{ props.item.status }}</td>
             <td>{{ props.item.created | moment("D MMM YYYY, HH:mm") }}</td>
+            <td>
+              <router-link
+                :to="{ name: 'jobDetails', params: { id: props.item.id } }"
+                class="link"
+                ><v-btn color="primary" flat>Details</v-btn></router-link
+              >
+            </td>
           </template>
         </v-data-table>
       </v-flex>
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -17,6 +17,7 @@
     "material-design-icons-iconfont": "^3.0.3",
     "roboto-fontface": "*",
     "vue": "^2.6.6",
+    "vue-moment": "^4.0.0",
```
ChristianLieven: This is super handy!

src/components/Jobs.vue
```diff
@@ -0,0 +1,110 @@
...
+      if (this.areAllJobsFinished) {
+        clearInterval(this.timerId);
+      }
+    }
+  },
+  computed: {
+    jobs() {
+      return this.$store.state.jobs.jobs;
+    },
+    ...mapGetters({
```
Midnighter: Awesome, I'd like to see this used more, rather than direct access from a component's template to `$store`. Also note that the `map*` functions accept the namespace argument if you ever want to map multiple, for example,
```js
...mapGetters("session", {
  x1:
  x2:
  x3:
})
```

:slightly_smiling_face:

src/components/Jobs.vue
```diff
@@ -0,0 +1,110 @@
...
+      descending: true
+    },
+    areAllJobsFinished: true,
+    timerId: 0
+  }),
+  methods: {
+    fetchJobs() {
+      this.areAllJobsFinished = true;
+      this.$store.dispatch("jobs/fetchJobs");
+      for (let i = 0; i < this.jobs.length; i++) {
```
Midnighter: I would change this whole rest of the method. You don't really use `this.areAllJobsFinished` anywhere else in the component. So it doesn't need to be in the component's state. Simply check if you need to clear the interval like so:

```js
if (this.jobs.some(job => {
  return (job.status === "STARTED" || job.status === "PENDING");
})) {
  clearInterval(this.timerId);
}
```

src/components/JobDetails.vue
```diff
@@ -0,0 +1,11 @@
+<template
```
Midnighter: This is still to come?

annasirunian: Yes, in the next PR

src/components/Jobs.vue
annasirunian: Thanks! :)

src/components/Jobs.vue
annasirunian: Thanks for the great advice! I just need the opposite check, so I used every():

```
      if (
        this.jobs.every(job => {
          return job.status !== "STARTED" && job.status !== "PENDING";
        })
      ) {
        clearInterval(this.timerId);
      }
```

Midnighter:
> Hmm, I'm thinking now that the logic to update jobs should be part of the jobs store, not the component. The component is only concerned with presenting the data as retrieved from the store. We might as well update the jobs in the background regardless of whether the jobs component is open or not.

On one hand that makes sense to me, on the other hand a user may never navigate to the jobs list so then it seems it's only needed when on the component.

src/components/Jobs.vue
```diff
@@ -0,0 +1,110 @@
...
+      </v-flex>
+    </v-layout>
+  </v-container>
+</template>
+
+<script lang="ts">
+import Vue from "vue";
+import { mapGetters } from "vuex";
+
+Vue.use(require("vue-moment"));
```
kvikshaug: Nice addition - let's add it globally instead?

annasirunian: Of course, will do it

kvikshaug:
> on the other hand a user may never navigate to the jobs list so then it seems it's only needed when on the component.

Data that's available through the store should be assumed to be up-to-date and usable by any component. Say for example we wanted to add an indicator to the main toolbar with the number of jobs in progress. Then that could give stale information depending on whether the user has the jobs component open or not.

src/components/Jobs.vue
Midnighter: Whoops, my bad. Although with @kvikshaug argument below. We probably don't need this method at all any longer :(

If we did use it, stick to `some` but negate the entire result:

```js
if (!this.jobs.some(job => {
  return (job.status === "STARTED" || job.status === "PENDING");
})) {
  clearInterval(this.timerId);
}
```

annasirunian:
> I don't understand yet when and where the polling will be triggered but the code itself looks good.

It will be triggered initially when user opens the platform, there could be running jobs (fetchJobs action is responsible for the whole logic now) and also when user starts a new job (we need to dispatch fetchJobs action in Design component later), and there is isLoading status in jobs store to prevent sending multiple requests at the same time if we are going to dispatch fetchJobs action from multiple components

PR #13, opened 2019-05-01
Chore guard routes

Midnighter:
1. This is an example of how routes can be protected based on authorization state.
2. Also replace the history of a user upon logout.

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -1,6 +1,6 @@
 import Vue from "vue";
 import Router from "vue-router";
-
+import store from "@/store";
 import sessionStore from "@/store/modules/session";
 import Home from "./components/Home.vue";
 import Designs from "./components/Designs.vue";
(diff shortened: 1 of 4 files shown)


src/router.ts
```diff
@@ -1,11 +1,21 @@
 import Vue from "vue";
 import Router from "vue-router";
+
```
kvikshaug: I like separating the imports like this, but I don't think it makes sense to just do it ad-hoc unless we can enforce the rules with prettier/eslint/other tools. At the very least, if you want to introduce this, you should update all other imports correspondingly to stay consistent.

Midnighter: The closest thing I could find was https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/newline-after-import.md but it doesn't split them. I can either revert this change or split all imports manually in a separate PR.

Midnighter:
The only problem I couldn't solve is that it would be nice to somehow display a banner that the user needs to log in to access a denied URL.

kvikshaug:
You could use a flag in the root store and put the template in the root app component, like we do for login messages, fetch errors etc. Note the computed property needs a setter for the vuetify component to update the `v-model` binding (if applicable).

PR #15, opened 2019-05-02
Some more interactive map features

kvikshaug:
- Split components (added `Card` and `Escher` sub-components)
- Initializing and loading map now displays loader as overlay
- Add `onData` getter to the relevant data stores. Allows callers to pass a function that waits for a potential fetch request to complete before running the passed function.
- By using the above getters, timing issues with choosing preferred map and adding a card with default values are now resolved (also when navigating directly to the interactive map)
- Some card controls are implemented (add new wild type card, remove card, select by clicking title) - lots still TBD.

### DIFF
--- a/src/components/InteractiveMap/Escher.vue
+++ b/src/components/InteractiveMap/Escher.vue
@@ -1,0 +1,143 @@
+<template>
+  <div class="escher-container fill-height">
+    <v-container
+      fluid
+      fill-height
+      class="overlay"
+      v-if="initializingEscher !== null || isLoadingMap"
+    >
+      <v-layout align-center justify-center>
+        <v-progress-circular
+          indeterminate
+          size="40"
+          :width="2"
+          class="mr-2"
+          color="white"
+        ></v-progress-circular>
+        <p class="display-1 white--text mb-0">
+          <span v-if="initializingEscher !== null">Initializing Escher...</span>
+          <span v-else-if="isLoadingMap">Loading map...</span>
+        </p>
+      </v-layout>
+    </v-container>
+    <div ref="escher" class="fill-height"></div>
+  </div>
+</template>
+
+<script lang="ts">
+import Vue from "vue";
+/// <reference path="@/types/escher.d.ts" />
+import * as escher from "@dd-decaf/escher";
+
+export default Vue.extend({
+  name: "Escher",
+  props: ["mapData", "fluxDistribution"],
+  data: () => ({
+    escherBuilder: null,
+    initializingEscher: null,
+    isLoadingMap: false
+  }),
+  watch: {
+    mapData(value) {
+      const loadMap = () => {
+        this.isLoadingMap = true;
+        // Delay the actual loading of the map to give Vue time to update the
+        // interface (after setting isLoadingMap to true). Otherwise, Escher
+        // would freeze execution until it's complete and the user would never
+        // actually see the loading message.
+        setTimeout(() => {
+          this.escherBuilder.load_map(value);
+          this.isLoadingMap = false;
+        }, 10);
+      };
+
+      // Wait for escher to initialize before loading the map.
+      if (this.initializingEscher !== null) {
+        this.initializingEscher.then(loadMap);
+      } else {
+        loadMap();
+      }
+    },
+    fluxDistribution(value) {
+      this.escherBuilder.set_reaction_data(value);
+      this.escherBuilder._update_data(true, true);
+    }
+  },
+  methods: {
+    reactionState(id: string, type?: string) {
+      // TODO
+    },
+    handleKnockout(reactionId: string) {
+      // TODO
+    },
+    handleKnockoutGenes(reactionId: string) {
+      // TODO
+    },
+    handleSetAsObjective(reactionId: string) {
+      // TODO
+    },
+    handleChangeBounds(reactionId: string, lower: string, upper: string) {
+      // TODO
+    },
+    handleResetBounds(reactionId: string) {
+      // TODO
+    },
+    handleObjectiveDirection(reactionId: string) {
+      // TODO
+    }
+  },
+  mounted() {
+    this.initializingEscher = new Promise((resolve, reject) => {
+      this.escherBuilder = escher.Builder(null, null, null, this.$refs.escher, {
+        menu: "zoom",
+        scroll_behavior: "zoom",
+        fill_screen: false,
+        ignore_bootstrap: true,
+        never_ask_before_quit: true,
+        reaction_styles: ["color", "size", "text", "abs"],
+        identifiers_on_map: "bigg_id",
+        hide_all_labels: false,
+        hide_secondary_metabolites: false,
+        highlight_missing: true,
+        reaction_scale: [
+          { type: "min", color: "#A841D0", size: 20 },
+          { type: "Q1", color: "#868BB2", size: 20 },
+          { type: "Q3", color: "#6DBFB0", size: 20 },
+          { type: "max", color: "#54B151", size: 20 }
+        ],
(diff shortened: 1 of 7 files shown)


src/components/InteractiveMap/Escher.vue
```diff
@@ -0,0 +1,143 @@
...
+        scroll_behavior: "zoom",
+        fill_screen: false,
+        ignore_bootstrap: true,
+        never_ask_before_quit: true,
+        reaction_styles: ["color", "size", "text", "abs"],
+        identifiers_on_map: "bigg_id",
+        hide_all_labels: false,
+        hide_secondary_metabolites: false,
+        highlight_missing: true,
+        reaction_scale: [
```
Midnighter: I will create an issue about this. Zak has updated Escher (I don't know if our fork has those changes) such that we can have more control over the scale. For example, we can set the same scale across multiple cards.

src/components/InteractiveMap/InteractiveMap.vue
```diff
@@ -54,49 +36,14 @@
...
-              </v-flex>
-              <v-flex class="xs6 text-xs-right">
-                {{ card.growthRate }} <em>h<sup>-1</sup></em>
-              </v-flex>
-            </v-layout>
-          </v-card-title>
-        </v-card>
+        <Card
+          v-for="card in cards"
+          :key="cards.indexOf(card)"
```
Midnighter: Two issues here:

1. Finding the index of the card is unnecessary, you can get the index from the [v-for](https://vuejs.org/v2/guide/list.html)
```js
v-for="(item, index) in items"
```
2. I'm not 100% certain on this but I think using the index as key can lead to problems with components not being destroyed properly. Say you create new cards that then end up at the same index. I think this can be messy. I'll need to read up on that, though.

src/components/InteractiveMap/InteractiveMap.vue
```diff
@@ -54,49 +36,14 @@
...
+          :card="card"
+          :isLastCard="cards.length === 1"
```
Midnighter: This tests if it's the only card. Should the test be `index === cards.length - 1`?

src/components/InteractiveMap/InteractiveMap.vue
```diff
@@ -105,105 +52,125 @@
...
-        organism: { name: "E. coli" },
-        growthRate: 0.874
-      }
-    ]
+    fluxDistribution: null,
+    mapData: null,
+    cards: []
   }),
   methods: {
+    escherLoaded() {
```
Midnighter: Confusion on my part: Why is this a method and not a computed property?

src/components/InteractiveMap/InteractiveMap.vue
```diff
@@ -105,105 +52,125 @@
...
-    reactionState(id: string, type?: string) {
-      // TODO
-    },
-    handleKnockout(reactionId: string) {
-      // TODO
-    },
-    handleKnockoutGenes(reactionId: string) {
-      // TODO
+    addDefaultCard() {
+      const model = this.$store.state.models.models.find(
```
Midnighter: Still think this sort of thing could be easily avoided by making `$store.state.models` an object. I'd rather convert the object to an array for the table views if necessary.

src/components/InteractiveMap/InteractiveMap.vue
Midnighter: I see now. Can you rename this variable to `isOnlyCard` maybe?

src/components/InteractiveMap/InteractiveMap.vue
```diff
@@ -105,105 +52,125 @@
...
+        growthRate: null,
+        isSelected: false
+      };
+      this.cards.push(card);
+      this.selectCard(card);
+      this.simulate(card);
     },
-    handleChangeBounds(reactionId: string, lower: string, upper: string) {
-      // TODO
+    removeCard(card) {
```
Midnighter: I would pass the index of the card to the card component itself. Then when triggering `removeCard` from the card component you pass in the index directly as an argument.

src/components/InteractiveMap/InteractiveMap.vue
```diff
@@ -105,105 +52,125 @@
...
+      this.cards.splice(this.cards.indexOf(card), 1);
     },
-    handleResetBounds(reactionId: string) {
-      // TODO
+    selectCard(card) {
```
Midnighter: How about instead of only having a property on the card component, you introduce a reference to the selected card in this component?

```js
this.selectedCard.isSelected = false;
this.selectedCard = card;
card.isSelected = true;
```

src/components/InteractiveMap/InteractiveMap.vue
```diff
@@ -105,105 +52,125 @@
...
+      this.cards.forEach(card => {
+        card.isSelected = false;
+      });
+      card.isSelected = true;
     },
-    handleObjectiveDirection(reactionId: string) {
-      // TODO
+    simulate(card) {
+      if (card.model === null) {
```
Midnighter: If necessary you can change the comparison to below in order to capture both `null` and `undefined`.

```suggestion
      if (card.model == null) {
```

src/components/InteractiveMap/InteractiveMap.vue
```diff
@@ -105,105 +52,125 @@
...
+        .then(response => {
+          card.isSimulating = false;
+        });
     }
   },
   computed: {
     maps() {
       return this.$store.state.maps.maps;
+    },
+    selectedCard() {
```
Midnighter: With my proposed change above this would simply be coming from local state (data) `this.selectedCard`.

src/components/InteractiveMap/Escher.vue
kvikshaug: Sounds good

src/components/InteractiveMap/InteractiveMap.vue
kvikshaug: Yes, this was intended temporarily. I'm not sure what to use for key, maybe we need to generate uuids for the cards since they don't necessarily always have a design primary key. I'll add a comment about this

src/components/InteractiveMap/InteractiveMap.vue
kvikshaug: This is an event listener that runs when escher is initialized. It calls the `addDefaultCard` method - but makes sure to do it only when models are available in the store (hence passing it through `$store.getters["models/onData"]`).

It does not return any data, so it's not a computed property.

src/components/InteractiveMap/InteractiveMap.vue
kvikshaug: What do you refer to by "this sort of thing" that you wish to avoid? You need to handle the case of the object missing either way, either through `find` + check for `undefined` or use `key in models`. Which one doesn't matter to me, but I don't see a reason to change the datastructure to differ from what's returned from the APIs, i.e., an ordered list of objects with PK as a field in each row.

src/components/InteractiveMap/InteractiveMap.vue
kvikshaug: That's a good idea :+1:

src/components/InteractiveMap/InteractiveMap.vue
kvikshaug: I think I'll keep this consistent with your suggestion of storing a reference to the selected card below, so handle and pass object references, not indices.

src/components/InteractiveMap/InteractiveMap.vue
Midnighter: It matters to me so I guess I'll make another PR eventually :)

PR #16, opened 2019-05-02
feat: add Privacy Policy and Terms in footer

phantomas1234:
image
image


phantomas1234:
Just grabbed generic policy templates from the internet, so still need to determine the right content but does the footer work for everyone? I check that on mobile it collapses automatically so it doesn't get in the way.

src/App.vue
```diff
@@ -176,7 +176,14 @@
       <v-content>
         <router-view />
       </v-content>
-      <v-footer app></v-footer>
+      <v-footer app justify-center dark color="primary" height="auto">
```
ChristianLieven: I think you don't need to add the dark attribute here. As far as I understand it is only supposed to be applied when the app has a defined dark theme.

https://vuetifyjs.com/en/framework/theme#light-and-dark

Midnighter:
How about putting an "About" in the nav menu or putting these elements at the bottom of the menu? I feel like the footer is taking away space unecessarily.

src/components/PrivacyPolicy.vue
```diff
@@ -0,0 +1,68 @@
...
+        <p>By continuing to navigate our website without changing your cookie settings, you hereby acknowledge and agree to Technical University of Denmark's use of cookies.</p>
+
+        <h2>Privacy Policy Changes</h2>
+        <p>Although most changes are likely to be minor, Technical University of Denmark may change its Privacy Policy from time to time, and in Technical University of Denmark's sole discretion. Technical University of Denmark encourages visitors to frequently check this page for any changes to its Privacy Policy. Your continued use of this site after any change in this Privacy Policy will constitute your acceptance of such change.</p>
+
+        <h2></h2>
+        <p></p>
+
+        <h2>Credit & Contact Information</h2>
```
kvikshaug:
```suggestion
        <h2>Credit &amp; Contact Information</h2>
```

src/components/PrivacyPolicy.vue
```diff
@@ -0,0 +1,68 @@
...
+        <p>By continuing to navigate our website without changing your cookie settings, you hereby acknowledge and agree to Technical University of Denmark's use of cookies.</p>
+
+        <h2>Privacy Policy Changes</h2>
+        <p>Although most changes are likely to be minor, Technical University of Denmark may change its Privacy Policy from time to time, and in Technical University of Denmark's sole discretion. Technical University of Denmark encourages visitors to frequently check this page for any changes to its Privacy Policy. Your continued use of this site after any change in this Privacy Policy will constitute your acceptance of such change.</p>
+
+        <h2></h2>
+        <p></p>
```
kvikshaug: These can be removed

Midnighter:
+1 for kebab case also for `interactive-map`.

src/App.vue
phantomas1234: The toolbar is also using `dark` https://github.com/DD-DeCaF/caffeine-vue/blob/697d999f31a047e93e721f20f21fe695de72fb14/src/App.vue#L4 otherwise the font is black. I also feel like it should be the job of vuetify to determine a font color that has enough contrast but it doesn't.

src/components/PrivacyPolicy.vue
phantomas1234: > I do like kebab-case in the URLs, but it's inconsistent with `/interactiveMap`. Should we change that to `/interactive-map` then?

Yes. All routes should be overhauled. `/pathways` should be `design` etc.

phantomas1234:
> How about putting an "About" in the nav menu or putting these elements at the bottom of the menu? I feel like the footer is taking away space unecessarily.

I agree, and found this example here 
image
Will change tomorrow.

src/App.vue
ChristianLieven: When @kvikshaug set up the skeleton app we didn't spend a lot of time understanding the theming options in Vuetify. I'd be keen revisit that once we've progressed to a point where the vue rebuild is legit :D

I think it is OK to keep the `dark` for now but, if we wanted solid theming we have to do much of it ourselves. Vue's defaults are rudimentary at best and after some testing, I can't really say that it automatically sets the font color for you.

src/components/TermsOfService.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+  <v-container>
+    <v-layout justify-center>
+      <v-flex md6>
+        <h2>DD-DeCaF Terms of Service</h2>
```
kvikshaug: We should probably use the [vuetify typography classes](https://vuetifyjs.com/en/framework/typography) here - but if not, at least use `h1` like on the other views for a consistent header size.

phantomas1234:
I think I am going to reopen this in a new PR because I messed up with some merge conflicts.

PR #19, opened 2019-05-02
chore: add an editorconfig for consistent layout

Midnighter:


.editorconfig
```diff
@@ -0,0 +1,18 @@
+root=true
+
+[*]
+charset=utf-8
+end_of_line=lf
+indent_style=space
+indent_size=2
+insert_final_newline=true
```
ChristianLieven: I don't really get why this is done. I see it everywhere but why are we enforcing it here but for instance makefiles and markdown don't require it?

Midnighter: Hmm, I used a template for this. To me a final newline makes sense also for markdown and make files. I'd be up for removing the rule there.

.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+trim_trailing_whitespace=true
+
+[Makefile]
```
kvikshaug: We don't use a makefile in this project, so you can remove this entire section.

PR #21, opened 2019-05-03
feat: wait for the token to be refreshed before concluding project creation

ChristianLieven:
Keeping the PRs small from now on. This is a change requested by @kvikshaug on #8 - https://github.com/DD-DeCaF/caffeine-vue/pull/8#pullrequestreview-233058310


ChristianLieven:
> LGTM. You're still planning to deal with error handling later? If the refresh token failed, I guess the user would now be left with a loading screen that never disappears.
> 
> I see that the build doesn't fail on the lint warnings you got - I'll suggest that in a separate PR.

Cheers! Yeah, I think handling the errors centrally makes sense. Many of the API related errors are similar between the shared components and the CRUD views.

PR #24, opened 2019-05-03
refactor: initialize store from local storage before creating Vue app

Midnighter:
* Route guard waits for local storage reading to have happened
* Add gaurds to further routes

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -14,11 +14,30 @@ import NotFound from "./components/NotFound.vue";
 Vue.use(Router);
 
 const authGuard = (to, from, next) => {
-  if (sessionStore.state.isAuthenticated) {
-    next();
+  // Check if role was loaded from local storage. This happens before App
+  // creation.
+  if (sessionStore.state.userRole === null) {
+    // Watch the role value for changes so we know when loading has happened.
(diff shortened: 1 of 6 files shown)


src/store/index.ts
```diff
@@ -65,5 +65,8 @@ export default new Vuex.Store({
       dispatch("organisms/fetchOrganisms");
       dispatch("projects/fetchProjects");
     }
-  }
+  },
+  strict: process.env.NODE_ENV !== "production"
```
kvikshaug: :+1:

src/router.ts
```diff
@@ -13,11 +13,30 @@ import Models from "./components/Models.vue";
...
   } else {
-    store.commit("setUnauthorizedError", to.path);
-    next({ name: "home" });
+    // Loading has happened, we proceed normally.
+    if (sessionStore.state.isAuthenticated) {
+      next();
+    } else {
+      store.commit("setUnauthorizedError", to.path);
+      next({ name: "home" });
+    }
```
kvikshaug: Consider de-duplicating the lines in this block by defining a function that you call both from the watcher above and this else clause (not relevant if we refactor to avoid asynchronicity).

PR #25, opened 2019-05-03
Job detail

annasirunian:
Header and status for the job details table:

- When job is running:
image

- When job was failed:
image


src/components/Jobs/JobDetails.vue
```diff
@@ -0,0 +1,119 @@
...
+              <v-layout align-center justify-end row>
+                <h2 class="ma-3">
+                  <router-link to="/jobs" class="link">Jobs</router-link> / #{{
+                    jobId
+                  }}
+                </h2>
+                <v-spacer></v-spacer>
+                <v-list-tile class="ma-3">
+                  <v-icon class="mr-2">bug_report</v-icon>
+                  <div class="header-item">
```
kvikshaug: Consider using the predefined [vuetify typography](https://vuetifyjs.com/en/framework/typography) instead of defining custom classes

src/components/Jobs/JobDetails.vue
```diff
@@ -0,0 +1,119 @@
+<template>
+  <div>
+    <template v-if="job">
+      <v-container>
+        <v-layout column>
```
kvikshaug: The job-details header doesn't seem to break properly on mobile screens.

image

PR #28, opened 2019-05-06
fix: establish reactivity on models and maps component for all operations

ChristianLieven:
This should get rid of the bug where adding or editing a model did not immediately update the table of maps and model components.


src/components/Maps.vue
```diff
@@ -305,17 +308,20 @@ export default Vue.extend({
     },
     passModel(model) {
       this.selectedModel = model;
+    },
+    passMap(model) {
+      console.log(model);
```
kvikshaug: This argument should probably be `map`. The `console.log` is to be replaced later?

src/components/Maps.vue
```diff
@@ -13,7 +13,10 @@
       v-model="isModelCreationDialogVisible"
       @returnObject="passModel"
     />
-    <NewMap v-model="isMapCreationDialogVisible" />
+    <NewMap 
+      v-model="isMapCreationDialogVisible" 
+      @returnObject="passMap"
```
kvikshaug: Event names are recommended [to be kebab-case](https://vuejs.org/v2/guide/components-custom-events.html#Event-Names).

src/components/Maps.vue
ChristianLieven: Nah this was a left over statement which I've since removed.

src/components/Maps.vue
ChristianLieven: True! Will fix it soon!

PR #29, opened 2019-05-07
feat: implement project component

ChristianLieven:


src/components/Projects.vue
```diff
@@ -0,0 +1,227 @@
...
+    <v-layout justify-center>
+      <v-flex md6>
+        <h1>Projects</h1>
+        <v-data-table
+          :headers="headers"
+          :items="availableProjects"
+          class="elevation-8"
+          :pagination.sync="pagination"
+        >
+          <template v-slot:items="props">
```
Midnighter: Not my favorite variable name since props and properties have a specific separate meaning for components.

ChristianLieven: I did not chose that as it is the default for https://vuetifyjs.com/en/components/data-tables#data-table

Midnighter: I see, cheers.

PR #30, opened 2019-05-08
Interactive map

kvikshaug:
- [x] Feat: Close sidepanel on map focus, fab button to show
- [x] Bugfix: Red X for gene/reaction knockouts in Escher is now correctly synced (thanks @annasirunian)
- [x] Feat: Model is now loaded in Escher
    - Reactions on map that are not in the model highlighted in red
    - Default bounds (from the model) are shown on reaction hover
- [x] Feat: Model modifications in card dialog implemented
    - Set objective (searches through model reactions)
    - Toggle objective direction
    - Add reaction from BiGG universal model (note: all reactions are assumed to be reversible)
    - Knock out reactions (searches through model reactions)
    - Knock out genes (searches through model genes)
    - Edit bounds (searches through model reactions, prefills fields with default bounds)

Remaining:

- [ ] Data-driven cards

### DIFF
--- a/src/bigg.ts
+++ b/src/bigg.ts
@@ -1,0 +1,43 @@
+import axios from "axios";
+import * as settings from "@/settings";
+
+export const lookupReaction = (reactionId: string) => {
+  // Return a reaction object with info from BiGG (returned in a delayed
+  // fashion)
+  const reaction: any = {
+    id: reactionId,
+    name: null,
+    reactionString: null
+  };
+  axios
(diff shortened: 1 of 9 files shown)


kvikshaug:
> LGTM, the dialog is great! The only small thing I would suggest is to highlight empty fields, for example, when you change the organism, you also need to change the model, but initially it's not clear, because there are loaders indicating that something is happening and it feels like the simulation is in progress

Good point - I'll also make sure loaders are only visible when a model is actually loading, and otherwise disable the related fields.

PR #37, opened 2019-05-13
Destructure slot props

kvikshaug:
Using object destructuring in the v-slot field, combined with renaming
Vuetifys `item` to the appropriate object name, makes the related
templates a little more readable.

Note: I wasn't able to make `expanded` work with destructuring in the
items slot for the designs table.

### DIFF
--- a/src/views/Designs.vue
+++ b/src/views/Designs.vue
@@ -63,7 +63,7 @@
                 <td>{{ props.item.design.gene_knockouts.length }}</td>
               </tr>
             </template>
-            <template v-slot:expand="props">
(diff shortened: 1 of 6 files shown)


kvikshaug:
> I'm not really sure what issues you ran across with `expanded`, is that something we should spend time on?

No I don't think it's important - it just means that this section https://github.com/DD-DeCaF/caffeine-vue/blob/63d15ad630ffcad78af8f9898dcfb8b800bc6534/src/views/Designs.vue#L43 will still be using `props` as slot prop object.

PR #43, opened 2019-05-15
feat: integrate sentry

ChristianLieven:
This integrates sentry with our Caffeine-Vue. 

Please bear this in mind:

```Vue Error Handling
Please note that if you enable this integration Vue internally will not call logError due to a currently know limitation see: GitHub Issue.
This means that errors occurring in the Vue renderer will not show up in the developer console.```
- https://docs.sentry.io/platforms/javascript/vue/?_ga=2.242518274.1462329972.1557918583-653711018.1550504433

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -16,15 +16,17 @@ import { initFromStorage } from "@/utils/startup";
 import * as Sentry from '@sentry/browser';
 import * as Integrations from '@sentry/integrations';
 
-Sentry.init({
-  dsn: 'https://44d530ac9cb14ad391a54689b2286a92@sentry.io/1460040',
-  integrations: [
-    new Integrations.Vue({
-      Vue,
-      attachProps: true,
-    }),
-  ],
-});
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/main.ts [hunk 1]
```diff
@@ -13,6 +13,18 @@ import LoaderDialog from "@/components/LoaderDialog.vue";
 import DeletionDialog from "@/components/DeletionDialog.vue";
 import FileUpload from "@/components/FileUpload.vue";
 import { initFromStorage } from "@/utils/startup";
+import * as Sentry from '@sentry/browser';
+import * as Integrations from '@sentry/integrations';
+
+Sentry.init({
```
Midnighter: Bearing your comment in mind, I think it's important to see errors in development and we don't want Sentry notifications for development anyway (at least not in the frontend). So I suggest to wrap this whole Sentry initialization:

```ts
if (process.env.NODE_ENV === "production") {
  Sentry.init(...);
}
```

ChristianLieven: Cheers!

PR #47, opened 2019-05-16
Project theming

ChristianLieven:


src/store/index.ts
```diff
@@ -36,7 +36,8 @@ export default new Vuex.Store({
     commonTooltipMessages: {
       unauthenticated: "Please log in or register to use this functionality!",
       publicData: "Public data can not be modified."
-    }
+    },
+    currentlyActiveProject: null
```
Midnighter: This makes more sense to me on the projects store (`src/store/project.ts`) since that is supposed to handle global state related to projects.

src/App.vue
```diff
@@ -238,9 +285,49 @@ export default Vue.extend({
...
+      this.$vuetify.theme.primary = colors.blue.base;
+      this.$store.commit("setCurrentlyActiveProject", null);
+    },
+    projectPrimaryColor(projectID) {
+      const sortedColors = Object.values(this.sensibleColors).sort();
+      return String(
+        sortedColors[projectID % Object.keys(this.sensibleColors).length]
+      );
+    },
+    camelToKebab(string) {
```
Midnighter: According to WebStorm, this method is not used anywhere. Are you planning to add functionality?

src/App.vue
```diff
@@ -238,9 +285,49 @@ export default Vue.extend({
...
+            if (shade.includes("base") || shade.includes("darken2")) {
+              obj[String(color + "_" + shade)] = colors[color][shade];
+            }
+          }
+        }
+      }
+      return obj;
+    }
+  },
+  methods: {
```
Midnighter: I was going to suggest to make most of these computed properties but now I think it would be better to move this functionality to the projects store.

src/store/index.ts
ChristianLieven: That's fair, I've only placed it there as this is where I was told it was managed in the previous platform.

src/App.vue
ChristianLieven: Oops no, this is a left-over from a previous commit. I'll remove it.

PR #48, opened 2019-05-16
Move cards to the store

kvikshaug:
As part of the designs/jobs visualizations.


src/views/InteractiveMap/Card.vue
```diff
@@ -181,27 +181,46 @@ export default Vue.extend({
       immediate: true,
       handler() {
         // Reset all modifications when the selected model changes.
-        this.card.reactionAdditions = [];
-        this.card.reactionKnockouts = [];
-        this.card.geneKnockouts = [];
-        this.card.editedBounds = [];
+        this.$store.commit("interactiveMap/updateCard", {
```
Midnighter: I would map this mutation into a local method as described in https://vuex.vuejs.org/guide/mutations.html#committing-mutations-in-components

src/store/modules/interactiveMap.ts
```diff
@@ -0,0 +1,253 @@
...
+    addCard(state, card) {
+      state.cards.push(card);
+    },
+    removeCard(state, card) {
+      state.cards.splice(state.cards.indexOf(card), 1);
+    },
+    updateCard(state, { uuid, props }) {
+      const card = state.cards.find(c => c.uuid === uuid);
+      Object.keys(props).forEach(key => {
+        card[key] = props[key];
```
lukaw3d: I don't think Vuex can detect these (and other) deep object changes (e.g. to re-run dependent `computed` properties on components)

Recommendation here: https://vuex.vuejs.org/guide/mutations.html#mutations-follow-vue-s-reactivity-rules

kvikshaug: That's indeed the case when you add new properties to an object. Here we're just modifying existing properties though, and that should still be detected by deep watchers. Regardless, it's probably good practice to just use `Vue.set` in general, so I'll update this - thanks.

src/views/InteractiveMap/Card.vue
kvikshaug: That's a nice improvement, thanks :+1:

PR #49, opened 2019-05-19
Added reactions

annasirunian:
Done:
1. Request full information from BIGG on added reactions when fetching designs
2. Add progress slot for designs and jobs data tables indicating that data is fetching

### DIFF
--- a/src/store/modules/designs.ts
+++ b/src/store/modules/designs.ts
@@ -14,6 +14,7 @@ export interface DesignItem {
     gene_knockouts: string[];
     reaction_knockins: string[];
     reaction_knockouts: string[];
+    added_reactions?: Object[];
   };
   id: number;
   model_id: number;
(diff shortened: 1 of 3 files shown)


src/store/modules/designs.ts
```diff
@@ -40,7 +41,73 @@ export default {
...
+                      upperBound: 1000,
+                      gene_reaction_rule: "",
+                      metabolites: {
+                        ...r.data.metabolites.map(m => ({
+                          [`${m.bigg_id}_${
+                            m.compartment_bigg_id
+                          }`]: m.stoichiometrys
+                        }))
+                      },
+                      metabolites_to_add: r.data.metabolites.map(m => ({
```
kvikshaug: We shouldn't need this now - it'd be enough to add metabolites with the interface from the interactive map store, then this will be automatically mapped to escher/cobra format when adding the reaction.

src/store/modules/designs.ts
```diff
@@ -40,7 +41,73 @@ export default {
...
+                .then(response => {
+                  let addedReactions: any = [];
+                  response.forEach(r => {
+                    const reaction = {
+                      id: r.data.bigg_id,
+                      name: r.data.name,
+                      reactionString: r.data.reaction_string,
+                      lowerBound: -1000,
+                      upperBound: 1000,
+                      gene_reaction_rule: "",
```
kvikshaug: This empty attribute isn't required anymore.

src/store/modules/designs.ts
```diff
@@ -14,6 +14,7 @@ export interface DesignItem {
     gene_knockouts: string[];
     reaction_knockins: string[];
     reaction_knockouts: string[];
+    added_reactions?: Object[];
```
kvikshaug: Hmm, I think we should overwrite `reaction_knockins` instead of adding a new property here.

PR #58, opened 2019-05-22
Visualize designs

kvikshaug:
In order to facilitate this, a couple extra features were necessary:
- Models are now not stored on the cards, but purely in the store
- Lazy loader action `withFullModel` in the model store lets callers guarantee that full model is available
- When adding reactions, model is now not directly modified, but a computed property based on the original model and card modifications
- Some escher sync issues are resolved

With this in place, the Designs component is now able to commit a card and redirect to the interactive map, visualizing the selected design(s).

### DIFF
--- a/src/store/modules/models.ts
+++ b/src/store/modules/models.ts
@@ -10,6 +10,8 @@ export interface ModelItem {
   project_id: number;
   preferred_id: number;
   default_biomass_reaction: string;
+  // Note that `model_serialized` will not be retrieved before requested through
+  // the `withFullModel` action.
   model_serialized?: object;
 }
 
@@ -23,11 +25,18 @@ export default {
   namespaced: true,
   state: {
     models: [],
-    modelsPromise: null
+    modelsPromise: null,
+    // `fullModelPromises` contains the promises to request the full model,
+    // keyed by model id. See the `withFullModel` action.
+    fullModelPromises: {}
   },
   mutations: {
     setModels(state, models: ModelItem[]) {
       state.models = models;
+    },
+    setFullModel(state, model) {
+      const index = state.models.findIndex(m => m.id === model.id);
+      Vue.set(state.models, index, model);
     },
     editModel(state, payload: any) {
       Vue.set(state.models, payload.index, payload.item);
@@ -40,6 +49,9 @@ export default {
     },
     setModelsPromise(state, modelsPromise) {
       state.modelsPromise = modelsPromise;
+    },
+    setFullModelPromise(state, { modelId, promise }) {
+      Vue.set(state.fullModelPromises, modelId, promise);
     }
   },
   actions: {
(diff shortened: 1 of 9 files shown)


src/views/Designs.vue
```diff
@@ -81,9 +87,9 @@
                     :size="15"
                   ></v-progress-circular>
                 </td>
-                <td>{{ props.item.design.reaction_knockins.length }}</td>
-                <td>{{ props.item.design.reaction_knockouts.length }}</td>
-                <td>{{ props.item.design.gene_knockouts.length }}</td>
+                <td>{{ props.item.design.reactionKnockins.length }}</td>
+                <td>{{ props.item.design.reactionKnockouts.length }}</td>
```
Midnighter: Apart from the `props.item` which is required by the Vuetify template, you are tightly coupling this component with the design implementation (see, for example, [this brief article](https://www.martinfowler.com/ieeeSoftware/coupling.pdf)). Concretely, it'd be better to have a property `numReactionKnockouts` on `design` so that, as an example, it'd be easy to change the actual structure from an array to an object in future without breaking a ton of code.

src/views/Designs.vue
```diff
@@ -308,6 +315,46 @@ export default Vue.extend({
...
+          conditionData: null,
+          conditionWarnings: [],
+          conditionErrors: [],
+          // General simulation fields
+          isSimulating: false,
+          hasSimulationError: false,
+          growthRate: null,
+          fluxes: null
+        };
+        // Make sure the full model is available before adding the card.
```
Midnighter: You say this is necessary for Escher? Otherwise it'd be neat for the browser to start drawing the interactive map component while still fetching the full model and then initializing the Escher map when a model is ready.

src/views/Designs.vue
Midnighter: Actually that article is not relevant to this case so much (other than talking about coupling). Maybe these are better resources (although I'm sure you know what I mean):

* https://blogs.msdn.microsoft.com/zainnab/2011/05/25/code-metrics-class-coupling/
* https://en.wikipedia.org/wiki/Coupling_%28computer_programming%29

src/views/Designs.vue
```diff
@@ -81,9 +87,9 @@
                     :size="15"
                   ></v-progress-circular>
                 </td>
-                <td>{{ props.item.design.reaction_knockins.length }}</td>
-                <td>{{ props.item.design.reaction_knockouts.length }}</td>
-                <td>{{ props.item.design.gene_knockouts.length }}</td>
+                <td>{{ props.item.design.reactionKnockins.length }}</td>
```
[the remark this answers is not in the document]
kvikshaug: It's not possible unfortunately because you need access to `props.expanded` here (or at least I couldn't figure out a way to achieve that).

src/store/modules/models.ts
```diff
@@ -40,19 +49,53 @@ export default {
...
+            .then(response => {
+              commit("setFullModel", response.data);
+              resolve(response.data);
+            })
+            .catch(error => {
+              commit("setFetchError", error, { root: true });
+              reject(error);
+            });
+        });
+        commit("setFullModelPromise", { modelId, promise });
```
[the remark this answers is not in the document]
kvikshaug: Yes indeed! Thanks.

src/store/modules/models.ts
```diff
@@ -40,19 +49,53 @@ export default {
...
         .get(`${settings.apis.modelStorage}/models`)
         .then((response: AxiosResponse<ModelItem[]>) => {
-          commit("setModels", response.data);
+          // Store the models, but initialize the missing `model_serialized`
+          // field to null.
+          commit(
+            "setModels",
+            response.data.map(model => ({
+              ...model,
+              model_serialized: null
```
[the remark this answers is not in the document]
kvikshaug: Yes, it's still currently consistently `null`. Your point about `null` mapping to JSON differently than `undefined` made me reconsider, and I'm now a bit more reluctant to make a PR to refactor all usage.

src/views/Designs.vue
kvikshaug: This wasn't introduced in this PR, I think it was in #7, maybe better to move the discussion there.

src/views/Designs.vue
kvikshaug: I absolutely agree. The main reason to delay the navigation is because there's logic in the interactive map component that needs to know whether there are already cards in store (see comments on [this commit](https://github.com/DD-DeCaF/caffeine-vue/commit/7ad284bd17b49ce7ba4d4fc37626afdf328858a3)).

If we don't wait, we'd need different logic to select the correct card, and make sure not to add a default since we're sharing designs from another component. I guess that should be possible somehow?

src/views/Designs.vue
Midnighter: Was meant as a general remark, my bad, probably better to start a discussion issue.

src/views/Designs.vue
Midnighter: I'll think about it :slightly_smiling_face:

PR #60, opened 2019-05-22
Visualize jobs

kvikshaug:


src/views/Jobs/JobResultsTable.vue
```diff
@@ -708,6 +751,98 @@ export default Vue.extend({
...
+            method: "pfba", // TODO - should this be default?
+            dataDriven: false,
+            // Design card fields
+            objective: {
+              reaction: null,
+              maximize: true
+            },
+            reactionAdditions: addedReactions,
+            reactionKnockouts: [],
+            geneKnockouts: [],
```
annasirunian: Now I realized that we missed the knockout part. We also have knockouts in the prediction and if method is opt_gene than those are gene knockouts, otherwise - reaction

PR #64, opened 2019-05-23
fix: min max attributes on sliders were not rounded

ChristianLieven:
Fixes: https://app.zenhub.com/workspaces/caffeine-platform-59dd1cab2fbdf4155c45e6e9/issues/dd-decaf/scrum/911

Potential follow up:
https://app.zenhub.com/workspaces/caffeine-platform-59dd1cab2fbdf4155c45e6e9/issues/dd-decaf/scrum/916

### DIFF
--- a/src/views/Jobs/JobResultsTable.vue
+++ b/src/views/Jobs/JobResultsTable.vue
@@ -167,10 +167,10 @@
                   <v-range-slider
                     @click.stop
                     v-model="filters.fitness"
-                    :min="range.fitness[0]"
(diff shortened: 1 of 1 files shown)


ChristianLieven:
> To how many places does Vue round floats by default? I think something like 3 decimals 1 could work?

We're not using native Vue behaviour here. The `round` filter I'm using is defined on line 585 as:

```
 round: value => {
      return value >= 0.005 || value === 0
        ? +value.toFixed(2)
        : value.toExponential(1);
    }
```

PR #69, opened 2019-05-24
Distinguish sentry prod/staging

kvikshaug:


kvikshaug:
> Is that an error or will the sentry init silently fail and thus do the right thing for us?

I wasn't able to find the API docs for it, but it did silently fail during my testing for dev mode.

> Small point: Maybe it'd be more in line with current practices to create a variable in `settings.ts`.

Yes indeed! I'll add that.

PR #71, opened 2019-05-24
Save designs

kvikshaug:
- [x] Cards can, when modified, be saved
- [x] User must select active project for design to be saved; it will be saved in that project
- [x] Deleting designs removes cards that are related to the design from the store

### DIFF
--- a/src/store/modules/interactiveMap.ts
+++ b/src/store/modules/interactiveMap.ts
@@ -39,7 +39,7 @@ export interface Card {
   dataDriven: boolean;
   // Design card fields
   objective: {
-    reaction: Reaction;
+    reaction: Reaction | null;
     maximize: boolean;
   };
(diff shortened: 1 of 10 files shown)


src/store/modules/interactiveMap.ts
```diff
@@ -33,6 +33,9 @@ export interface Card {
   organism: OrganismItem;
   modelId: number;
   method: string;
+  // `modified` is true when the user has made any change to the model, and it
+  // can be saved. Not relevant for data-driven cards.
```
annasirunian: Not sure if user should be able to save the card when for example he changed the organism, didn't select the model and then closed the dialog, now we are getting 422

kvikshaug: Thanks, that shouldn't be possible. I'll add a fix for this

PR #77, opened 2019-05-27
Open dialog when adding new design or data driven card

annasirunian:
https://app.zenhub.com/workspaces/caffeine-platform-59dd1cab2fbdf4155c45e6e9/issues/dd-decaf/scrum/910

### DIFF
--- a/src/views/InteractiveMap/Card.vue
+++ b/src/views/InteractiveMap/Card.vue
@@ -376,6 +376,19 @@ export default Vue.extend({
       }
     }
   },
+  mounted() {
+    if (this.withDialog) {
+      this.isCardDialogVisible = true;
+    }
+  },
+  destroyed() {
+    this.updateCard({
+      uuid: this.card.uuid,
+      props: {
+        withDialog: false
+      }
+    });
+  },
   methods: {
     selectCard() {
       this.$emit("select-card", this.card);
@@ -492,19 +505,6 @@ export default Vue.extend({
       updateCard: "interactiveMap/updateCard",
       setModified: "interactiveMap/setModified"
     })
-  },
-  mounted() {
-    if (this.withDialog) {
-      this.isCardDialogVisible = true;
-    }
-  },
-  destroyed() {
-    this.updateCard({
-      uuid: this.card.uuid,
-      props: {
-        withDialog: false
-      }
-    });
   }
 });
 </script>
--- a/src/views/InteractiveMap/Card.vue
+++ b/src/views/InteractiveMap/Card.vue
@@ -216,7 +216,7 @@ export default Vue.extend({
       return value.toFixed(3);
     }
   },
-  props: ["card", "isOnlyCard", "isSelected"],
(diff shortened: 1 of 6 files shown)


src/views/InteractiveMap/InteractiveMap.vue
```diff
@@ -300,7 +305,10 @@ export default Vue.extend({
         isSimulating: false,
         hasSimulationError: false,
         growthRate: null,
-        fluxes: null
+        fluxes: null,
+        // When adding a card from the interactive map, the dialog should be opened
+        // Method always exists on the default card
```
kvikshaug: I don't think we should rely on the method behaviour, it's not really related. Rather just add a new `withDialog` parameter to the `addCard` method.

src/views/InteractiveMap/InteractiveMap.vue
```diff
@@ -300,7 +305,10 @@ export default Vue.extend({
         isSimulating: false,
         hasSimulationError: false,
         growthRate: null,
-        fluxes: null
+        fluxes: null,
+        // When adding a card from the interactive map, the dialog should be opened
```
kvikshaug: Good comment, but maybe it's event better to place it in the Card interface?

src/views/InteractiveMap/InteractiveMap.vue
```diff
@@ -90,6 +90,7 @@
           :card="card"
           :isSelected="card === selectedCard"
           :isOnlyCard="cards.length === 1"
+          :withDialog="card.withDialog"
```
kvikshaug: No need to pass this - it's already available to the component through `this.card.withDialog`

src/views/InteractiveMap/InteractiveMap.vue
```diff
@@ -255,6 +256,10 @@ export default Vue.extend({
           this.hasLoadMapError = true;
         });
     },
+    addEmptyCard(dataDriven) {
+      const name = dataDriven ? "Data driven" : "Design";
+      this.addCard(name, null, null, "", dataDriven);
```
kvikshaug: I think it's preferable to use `null` here for the method, for consistency, rather than empty string

src/views/InteractiveMap/Card.vue
```diff
@@ -309,6 +314,19 @@ export default Vue.extend({
       }
     }
   },
+  mounted() {
```
kvikshaug: These will need to be reordered when merged into devel, because of the new property order rule, so just remember to run linter before pushing

src/views/InteractiveMap/InteractiveMap.vue
kvikshaug: Since `method` can be unset now, it would be nice with a default text label in the card (something like "Not selected"), like we have for organism and model.

image

phantomas1234: Sorry for not thinking this through, I think method should be preselected as pfba.

PR #81, opened 2019-05-28
Job result table

annasirunian:
Done:
- in the expanded row manipulations should be sorted by score
- add tooltip on manipulation showing the score
- add warning snackbar when visualizing diff_fva method
- when visualizing job with knockouts, check if we have a response from bigg for the details, otherwise we need to display only id on the dialog

### DIFF
--- a/src/views/Jobs/JobResultsTable.vue
+++ b/src/views/Jobs/JobResultsTable.vue
@@ -879,6 +879,14 @@ export default Vue.extend({
                     uuid: card.uuid,
                     geneId: geneId
                   });
+                  this.$store.commit("interactiveMap/updateKnockoutReaction", {
+                    uuid: card.uuid,
+                    gene: {
+                      id: geneId,
+                      name: geneId,
+                      reactions: []
+                    }
+                  });
                 });
               } else if (
                 jobPrediction.method === "PathwayPredictor+DifferentialFVA" ||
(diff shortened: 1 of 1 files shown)


src/views/Jobs/JobResultsTable.vue
```diff
@@ -841,6 +879,14 @@ export default Vue.extend({
                     uuid: card.uuid,
                     geneId: geneId
                   });
+                  this.$store.commit("interactiveMap/updateKnockoutReaction", {
+                    uuid: card.uuid,
+                    gene: {
+                      id: geneId,
+                      name: geneId,
```
kvikshaug: Guess we should rather leave name empty (e.g. "N/A") than pretend the ID is the same as the name

src/views/Jobs/JobResultsTable.vue
```diff
@@ -851,6 +897,14 @@ export default Vue.extend({
                     uuid: card.uuid,
                     reactionId: reactionId
                   });
+                  this.$store.commit("interactiveMap/updateKnockoutReaction", {
+                    uuid: card.uuid,
+                    reaction: {
+                      id: reactionId,
+                      name: reactionId,
```
kvikshaug: As for genes - don't put the ID here

src/views/Jobs/JobResultsTable.vue
```diff
@@ -780,6 +810,14 @@ export default Vue.extend({
...
       } else {
         this.selected = [...this.filteredPathways];
+        if (
+          this.selected.some(
+            jobPrediction =>
+              jobPrediction.method === "PathwayPredictor+DifferentialFVA"
+          )
+        ) {
+          this.isDiffFvaChecked = true;
+        }
```
kvikshaug: Minor detail, but you could reuse the `onCheckboxChange` method here to avoid implementing the logic twice.

```
this.selected.forEach(jobPrediction => { this.onCheckboxChange(jobPrediction.method); });
```

PR #82, opened 2019-05-29
Create v-select wrapper that autoselects the only option

lukaw3d:


src/views/InteractiveMap/CardDialog.vue
```diff
@@ -14,7 +14,7 @@
               ></v-text-field>
             </v-flex>
             <v-flex xs12 md3>
-              <v-select
+              <AutoSelect
```
lukaw3d: Open to renaming suggestions. `v-select-auto-single`?

Midnighter: `AutoVSelect`?

src/components/AutoSelect.vue
```diff
@@ -0,0 +1,39 @@
...
+      this.autoselectOnlyOption(vselect);
+    });
+  },
+  methods: {
+    autoselectOnlyOption(vselect: VSelect) {
+      if (vselect.multiple) {
+        throw new Error("AutoSelect does not support multiple selections.");
+      }
+      const firstOption = vselect.allItems[0];
+      const isSelectionEmpty = vselect.selectedItems.length === 0;
```
Midnighter: I simply don't know: This is always guaranteed to be an array even though you explicitly don't want this for multi selects?

lukaw3d: Exactly, v-select's internals are generalized like this to support both modes

lukaw3d:
I'll go with v-select-auto

PR #83, opened 2019-05-29
fix: allow preferred map to be empty

lukaw3d:
This was broken since 59c22779e499eb4b6b221cb3cd9b86fdd0425bab


lukaw3d:
Ah, I only now noticed discussion at https://github.com/DD-DeCaF/caffeine-vue/pull/74#discussion_r287780537

PR #84, opened 2019-05-29
feat: enable devtools outside of production, ignoring configuration type

lukaw3d:


src/store/index.ts
```diff
@@ -78,7 +78,7 @@ export default new Vuex.Store({
       commit("setFetchError", true);
     }
   },
-  strict: process.env.NODE_ENV !== "production"
-  // Depends on https://github.com/vuejs/vuex/pull/1478 being released.
-  // devtools: process.env.NODE_ENV !== "production"
+  strict: process.env.NODE_ENV !== "production",
+  // Remove spread workaround after https://github.com/vuejs/vuex/pull/1478 is released.
+  ...{ devtools: process.env.NODE_ENV !== "production" }
```
lukaw3d: I'm not sure if you want this, but it works

PR #87, opened 2019-06-07
fix: remove old service workers

Midnighter:


src/main.ts
```diff
@@ -18,6 +18,17 @@ import { initFromStorage } from "@/utils/startup";
 import * as Sentry from "@sentry/browser";
 import * as Integrations from "@sentry/integrations";
 
+// Temporarily remove old service workers until we re-introduce them.
+if ("serviceWorker" in navigator) {
+  navigator.serviceWorker.getRegistrations().then(registrations => {
+    if (registrations.length > 0) {
```
kvikshaug: This check isn't strictly necessary

PR #88, opened 2019-06-11
Add arbitrary reaction

annasirunian:
todo:
- [x] autocompletetion of metabolites from metanetx service
- [x] add any arbitrary metabolite identifier

### DIFF
--- a/src/views/InteractiveMap/CardDialogDesign.vue
+++ b/src/views/InteractiveMap/CardDialogDesign.vue
@@ -163,6 +163,12 @@
       @change="addReaction"
     ></v-autocomplete>
 
+    <v-icon>add</v-icon>
+    <v-btn @click.stop="isReactionDialogVisible = true"
+      >Add arbitrary reaction
+      <ReactionDialog v-model="isReactionDialogVisible" :model="model"
+    /></v-btn>
+
     <v-autocomplete
       v-model="knockoutReactionItem"
       :items="reactions"
@@ -250,9 +256,13 @@
 import Vue from "vue";
 import axios from "axios";
 import * as settings from "@/utils/settings";
+import ReactionDialog from "@/views/InteractiveMap/ReactionDialog.vue";
 
 export default Vue.extend({
   name: "CardDialogDesign",
+  components: {
+    ReactionDialog
+  },
   props: ["card", "model", "modifications"],
   data: () => ({
     modificationsHeaders: [
@@ -274,6 +284,7 @@ export default Vue.extend({
     // Edit bounds
     editBoundsReaction: null,
     editBoundsValid: false,
+    isReactionDialogVisible: false,
     editBoundsReactionRule: value =>
       (value && value.id && value.id.length > 0) ||
       "Please specify the reaction",
--- a/src/views/InteractiveMap/CardDialogDesign.vue
+++ b/src/views/InteractiveMap/CardDialogDesign.vue
@@ -163,10 +163,11 @@
       @change="addReaction"
     ></v-autocomplete>
 
-    <v-icon>add</v-icon>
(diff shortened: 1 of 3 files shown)


src/views/InteractiveMap/ReactionDialog.vue
```diff
@@ -0,0 +1,401 @@
...
+      };
+
+      // Add the reaction to the card.
+      this.$store.commit("interactiveMap/addReaction", {
+        uuid: this.card.uuid,
+        reaction: reaction
+      });
+
+      this.$emit("simulate-card");
+      this.$refs.form.reset();
```
lukaw3d: You can now replace it with `this.clear()`

annasirunian: Done, thanks!

src/views/InteractiveMap/ReactionDialog.vue
```diff
@@ -0,0 +1,509 @@
...
+                  >
+                    <v-icon color="primary">delete</v-icon></v-btn
+                  ></v-layout
+                ></v-flex
+              >
+            </v-layout>
+          </div>
+        </v-layout>
+        <div class="body-2 mb-2">
+          Product:
```
kvikshaug: I'd say `Products` here since substrates is also pluralized

src/views/InteractiveMap/ReactionDialog.vue
```diff
@@ -0,0 +1,509 @@
...
+        stoichiometry: -m.stoichiometry
+      }));
+
+      const reaction: Reaction = {
+        id: this.reactionId,
+        name: this.reactionName,
+        reactionString: this.reactionString,
+        // Note: Assuming all reactions in the universal model are
+        // reversible, but this might not be the case. Could potentially use
+        // the reaction string to check reversibility.
```
kvikshaug: This comment is not relevant here, should be removed

src/views/InteractiveMap/ReactionDialog.vue
```diff
@@ -0,0 +1,509 @@
...
+        upperBound: this.upperBound,
+        metabolites: [...negativeSubstrates, ...this.products]
+          .filter(m => m.metabolite)
+          .map(m => ({
+            id: this.getMetaboliteId(m.metabolite.id, m.metabolite.compartment),
+            name: m.metabolite.name,
+            compartment: m.compartment,
+            stoichiometry: m.stoichiometry
+            // formula: null,
+            // annotation: null
```
kvikshaug: I think we can remove `formula` and `annotation` from here

src/views/InteractiveMap/ReactionDialog.vue
```diff
@@ -0,0 +1,509 @@
...
+      this.$emit("simulate-card");
+      this.clear();
+      this.showDialog = false;
+    },
+    setCompartmentFromMetabolite(metabolite, value) {
+      metabolite.compartment = value.compartment;
+    },
+    passMetabolite(metabolite) {
+      this.customMetabolites.push(metabolite);
+      let target = this.isSubstratesMetabolite
```
kvikshaug: Prefer `const` over `let` when possible

src/views/InteractiveMap/CardDialogDesign.vue
```diff
@@ -163,6 +163,17 @@
       @change="addReaction"
     ></v-autocomplete>
 
+    <v-icon :disabled="!model">add</v-icon>
+    <v-btn @click.stop="isReactionDialogVisible = true" :disabled="!model"
+      >Add arbitrary reaction
```
kvikshaug: Maybe the label could rather be something like "Define your own reaction" to make more sense to the user

PR #92, opened 2019-06-14
chore: update axios

kvikshaug:
Fixes https://nvd.nist.gov/vuln/detail/CVE-2019-10742


lukaw3d:
It wasn't really such a severe security issue for Axios use in browser

PR #93, opened 2019-06-18
Remove bigg search

kvikshaug:


src/views/InteractiveMap/CardDialogDesign.vue
```diff
@@ -434,55 +370,7 @@ export default Vue.extend({
     dontFilterByDisplayedText() {
```
lukaw3d: Also not needed anymore

src/views/InteractiveMap/CardDialogDesign.vue
```diff
@@ -370,43 +343,6 @@ export default Vue.extend({
...
-      }
-
-      this.isLoadingAddReaction = true;
-      axios
-        .get(
-          `${settings.apis.bigg}/search?query=${query}&search_type=reactions`
-        )
-        .then(response => {
-          // Wait for the full model to be able to filter out results that
-          // already exist in the model.
```
lukaw3d: We are missing this kind of filtering in mnx. Do we need it?

kvikshaug: I think it's fine to ignore it, it's not that important to filter them out. Maybe there could be some use case to add a reaction twice now with minor modifications for example.

PR #96, opened 2019-06-24
feat: add demo mode

phantomas1234:
A demo mode is needed to facilitate the deployment of Caffeine for evaluation purposes (see also https://github.com/dd-decaf/caffeine-bootstrap) into an environment without internet access.
Also, a script for generating a demo image of caffeine-vue and pushing it to Dockerhub has been included.

### DIFF
--- a/.env.demo
+++ b/.env.demo
@@ -1,0 +1,15 @@
+VUE_APP_ENVIRONMENT=demo
+# Backend endpoints
+VUE_APP_API=http://localhost:4200
+VUE_APP_MODEL_API=http://localhost:8004
+VUE_APP_MODEL_STORAGE_API=http://localhost:8005
+VUE_APP_WAREHOUSE_API=http://localhost:8006
+VUE_APP_MAPS_API=http://localhost:8002
+VUE_APP_IAM_API=http://localhost:8001
+VUE_APP_METABOLIC_NINJA_API=http://localhost:8003
+VUE_APP_DESIGN_STORAGE_API=http://localhost:8007
+VUE_APP_ID_MAPPER_API=http://localhost:8008
+VUE_APP_METANETX_API=http://localhost:8009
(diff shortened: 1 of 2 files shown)


.env.demo
```diff
@@ -0,0 +1,15 @@
+VUE_APP_ENVIRONMENT=demo
+# Backend endpoints
+VUE_APP_API=http://localhost:4200
```
kvikshaug: This can be removed

.env.demo
```diff
@@ -0,0 +1,15 @@
+VUE_APP_ENVIRONMENT=demo
```
kvikshaug: I don't think we'll accomodate for a separate `demo` environment, just use staging here.

.env.demo
```diff
@@ -0,0 +1,15 @@
...
+VUE_APP_MODEL_API=http://localhost:8004
+VUE_APP_MODEL_STORAGE_API=http://localhost:8005
+VUE_APP_WAREHOUSE_API=http://localhost:8006
+VUE_APP_MAPS_API=http://localhost:8002
+VUE_APP_IAM_API=http://localhost:8001
+VUE_APP_METABOLIC_NINJA_API=http://localhost:8003
+VUE_APP_DESIGN_STORAGE_API=http://localhost:8007
+VUE_APP_ID_MAPPER_API=http://localhost:8008
+VUE_APP_METANETX_API=http://localhost:8009
+VUE_APP_BIGG_API=https://api.dd-decaf.eu/bigg
```
kvikshaug: We don't need to repeat this here, it's already defined in `.env`.

scripts/demo.sh
```diff
@@ -0,0 +1,4 @@
+# Build demo image and push to dockerhub (you need to login to dockerhub first)
```
kvikshaug: Does this script need a license header?

phantomas1234: Let's do this for the whole codebase at once.

PR #99, opened 2019-07-01
Cofactor swap

annasirunian:
https://app.zenhub.com/workspaces/caffeine-platform-59dd1cab2fbdf4155c45e6e9/issues/dd-decaf/scrum/957

### DIFF
--- a/src/views/Jobs/JobResultsTable.vue
+++ b/src/views/Jobs/JobResultsTable.vue
@@ -422,9 +422,16 @@
                             >
                               {{ manipulation.id }}
                             </a>
-                            <div
-                              v-html="getSwapManipulation(manipulation)"
-                            ></div>
+                            <div>
+                              <div
+                                v-for="(swap, index) in getSwapManipulation(
+                                  manipulation
+                                )"
+                                :key="index"
+                              >
+                                {{ swap }}
+                              </div>
+                            </div>
                           </div>
                           <div
                             v-if="index >= 10"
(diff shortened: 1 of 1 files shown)


src/views/Jobs/JobResultsTable.vue
```diff
@@ -914,6 +972,89 @@ export default Vue.extend({
...
+                    const stoichiometry =
+                      reactionToAdd.metabolites[metaboliteId];
+                    stoichiometry < 0
+                      ? substrates.push(
+                          Math.abs(stoichiometry) + " " + metaboliteId
+                        )
+                      : products.push(stoichiometry + " " + metaboliteId);
+                    return {
+                      id: metaboliteId,
+                      // assuming that compartment is always "c"
```
kvikshaug: This comment could be a bit more clear. _Why_ are we assuming that? And should it be reconsidered later?

src/views/Jobs/JobResultsTable.vue
```diff
@@ -914,6 +972,89 @@ export default Vue.extend({
...
+                  const metabolites = Object.keys(
+                    reactionToAdd.metabolites
+                  ).map(metaboliteId => {
+                    const stoichiometry =
+                      reactionToAdd.metabolites[metaboliteId];
+                    stoichiometry < 0
+                      ? substrates.push(
+                          Math.abs(stoichiometry) + " " + metaboliteId
+                        )
+                      : products.push(stoichiometry + " " + metaboliteId);
```
lukaw3d: I don't think sideeffects in ternary operator are obvious when reading the code.
And also one scope higher, sideeffects inside array.map (instead of array.forEach)

src/views/Jobs/JobResultsTable.vue
annasirunian: Yes, it should be reconsidered later, but it will be a far future when we change it

src/views/Jobs/JobResultsTable.vue
annasirunian: Thanks! Changed that

PR #100, opened 2019-07-02
Require curly braces

kvikshaug:
This is a suggestion to add the [`curly`](https://eslint.org/docs/rules/curly) eslint rule to require curly braces for code clarity. For example, the following:

```
if (foo) foo++;
```

Must be rewritten as:

```
if (foo) {
    foo++;
}
```

### DIFF
--- a/src/views/InteractiveMap/CardDialogDesign.vue
+++ b/src/views/InteractiveMap/CardDialogDesign.vue
@@ -378,9 +378,15 @@ export default Vue.extend({
           const fullMetabolite = addReaction.mnxReaction.metabolites.find(
             ({ mnx_id }) => mnx_id === m.id
           );
-          if (!fullMetabolite) return m;
-          if (!fullMetabolite.annotation) return m;
-          if (!fullMetabolite.annotation.bigg) return m;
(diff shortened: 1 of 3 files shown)


src/views/InteractiveMap/CardDialogDesign.vue
```diff
@@ -378,9 +378,15 @@ export default Vue.extend({
           const fullMetabolite = addReaction.mnxReaction.metabolites.find(
             ({ mnx_id }) => mnx_id === m.id
           );
-          if (!fullMetabolite) return m;
-          if (!fullMetabolite.annotation) return m;
-          if (!fullMetabolite.annotation.bigg) return m;
```
lukaw3d: Oh, I see you didn't use this style elsewhere in code. How would most of you create reactionWithModelMetabolites? (you probably have a better way than reformatted version of my short guard clauses)

kvikshaug: Hmm, maybe there is better way I don't see - but I think the reformatted version, although also more verbose, at least is more readable, and I always value readability over conciseness.

PR #102, opened 2019-07-04
fix: undo reaction knockout

annasirunian:
While deleting added reaction modification, we are checking if that reaction was knocked out before to delete also that knockout. Previously the last reaction knockout was removed regardless of id.

### DIFF
--- a/src/views/InteractiveMap/CardDialogDesign.vue
+++ b/src/views/InteractiveMap/CardDialogDesign.vue
@@ -491,10 +491,15 @@ export default Vue.extend({
           reactionId: modification.id
         });
         // If the objective was set to the added reaction, clear it.
-        this.$store.commit("interactiveMap/setObjectiveReaction", {
-          uuid: this.card.uuid,
-          reaction: null
-        });
+        if (
+          this.card.objective.reaction &&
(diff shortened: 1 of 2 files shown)


lukaw3d:
Looking at other `findIndex + splice`, it looks like undoEditBounds will have the same issue

annasirunian:
@lukaw3d true, thank you, will fix that, also the objective is set to null after each deletion

src/views/InteractiveMap/CardDialogDesign.vue
```diff
@@ -491,10 +491,15 @@ export default Vue.extend({
...
         // If the objective was set to the added reaction, clear it.
-        this.$store.commit("interactiveMap/setObjectiveReaction", {
-          uuid: this.card.uuid,
-          reaction: null
-        });
+        if (
+          this.card.objective.reaction &&
+          this.card.objective.reaction.id === modification.id
+        ) {
+          this.$store.commit("interactiveMap/setObjectiveReaction", {
```
lukaw3d: It bothers me a bit that this mutation is different from others. Maybe create `interactiveMap/undoObjectiveReaction` and move `if id == id` inside, for consistency

PR #103, opened 2019-07-04
Map to bigg

annasirunian:
1. If id-mapper returned several bigg ids mapped to metanetx id of metabolite, take the one that is in the model
2. Remove compartment from metabolite id when adding reaction with swapped cofactors, so it will not be doubled later

### DIFF
--- a/src/views/Jobs/JobResultsTable.vue
+++ b/src/views/Jobs/JobResultsTable.vue
@@ -1049,7 +1049,10 @@ export default Vue.extend({
                   lowerBound: lowerBound,
                   upperBound: upperBound,
                   reactionString: reactionString,
-                  metabolites: metabolites
+                  metabolites: metabolites.map((metabolite) => ({
+                    ...metabolite,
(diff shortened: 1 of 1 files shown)


src/views/Jobs/JobResultsTable.vue
```diff
@@ -1115,6 +1115,12 @@ export default Vue.extend({
               const metabolite = { id, name, compartment, stoichiometry };
               if (biggIds.hasOwnProperty(mnxId)) {
                 metabolite.id = biggIds[mnxId][0];
+                // if id-mapper returned more than 1 mapped bigg id, take the one that is in the model
+                biggIds[mnxId].forEach(id => {
+                  if (metabolitesInModel.has(id + "_c")) {
+                    metabolite.id = id;
+                  }
```
lukaw3d: Simpler
```js
metabolite.id = biggIds[mnxId].find(id => metabolitesInModel.has(id + "_c")) || 
                biggIds[mnxId][0]
```

PR #107, opened 2019-07-08
v-select-extended and v-autocomplete-extended with autoselectOnlyOption

lukaw3d:
Removed `<v-select-auto>`.
Added `<v-select-extended>` and `<v-autocomplete-extended>`. They can be used everywhere, and they add support for optional attribute `autoselectOnlyOption`. Both should be easy to extend with more mixins

Added autoselectOnlyOption to inputs in:
- src/views/Design.vue
- src/views/InteractiveMap/CardDialogDataDriven.vue
- src/views/InteractiveMap/CardDialog.vue


lukaw3d:
Moved into a separate folder for a cleaner future

PR #108, opened 2019-07-08
Aerobic anaerobic

annasirunian:
https://app.zenhub.com/workspaces/caffeine-platform-59dd1cab2fbdf4155c45e6e9/issues/dd-decaf/scrum/928

If user chose anaerobic conditions for the job, then during the visualization we need to set lower bound of the `EX_o2_e` reaction to 0

### DIFF
--- a/src/views/Jobs/JobResultsTable.vue
+++ b/src/views/Jobs/JobResultsTable.vue
@@ -949,8 +949,20 @@ export default Vue.extend({
                   : null,
               showDiffFVAScore: false
             };
-            // Knockout genes or reactions depending on method
             this.$store.commit("interactiveMap/addCard", card);
+            // Apply anaerobic conditions if needed
+            if (!jobPrediction.aerobic) {
+              {
+                const payload = {
+                  uuid: card.uuid,
+                  reactionId: "EX_o2_e",
+                  lowerBound: 0,
+                  upperBound: 1000
+                };
+                this.$store.dispatch("interactiveMap/editBounds", payload);
+              }
+            }
+            // Knockout genes or reactions depending on method
             if (jobPrediction.method === "PathwayPredictor+OptGene") {
               jobPrediction.knockouts.forEach(geneId => {
                 this.$store.dispatch("interactiveMap/knockoutGene", {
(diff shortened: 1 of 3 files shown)


src/views/Jobs/JobResultsTable.vue
```diff
@@ -949,8 +949,20 @@ export default Vue.extend({
                   : null,
               showDiffFVAScore: false
             };
-            // Knockout genes or reactions depending on method
             this.$store.commit("interactiveMap/addCard", card);
+            // Apply anaerobic conditions if needed
+            if (!jobPrediction.aerobic) {
+              {
```
lukaw3d: Typo

src/views/Jobs/Jobs.vue
```diff
@@ -13,10 +13,34 @@
...
+            <td
+              v-if="
+                model(job.model_id) && organism(model(job.model_id).organism_id)
+              "
+            >
               {{ organism(model(job.model_id).organism_id).name }}
             </td>
-            <td>{{ model(job.model_id).name }}</td>
+            <td v-else>
+              <v-progress-circular
```
lukaw3d: Does it matter if this gets stuck loading forever? e.g. if user runs a job on a new organism, and then deletes that organism

(v-if above nicely fixes an error that blocks the whole page from loading in this example)

src/views/Jobs/JobResultsTable.vue
```diff
@@ -949,8 +949,20 @@ export default Vue.extend({
...
+                const payload = {
+                  uuid: card.uuid,
+                  reactionId: "EX_o2_e",
```
kvikshaug: What about the models where this might have a different identifier?

src/views/Jobs/Jobs.vue
annasirunian: Yes, there is a dependency issue that is in our todo list. We need to decide how to deal with that, maybe need to delete all dependencies (the same concerns designs when user deletes model that was used for it).

src/views/Jobs/JobResultsTable.vue
annasirunian: Will create an issue on that

PR #115, opened 2019-07-12
Frontend fva

ChristianLieven:
Removes obsolete logic associated with previous diffFVA response from the backend. 

The new response now looks like this:
```
{
  id: string;
  score: number;
  value: number;
  direction: ["up", "down", "invert"];
}
```
The new logic now simply constrains the model with the 'value' parameters from diffFVA and runs a simulation once:

```
direction === "up"

& value > 0
--> reaction.lower_bound = value

& value < 0
--> reaction.upper_bound = value
```

```
direction === "down"

& value > 0
--> reaction.upper_bound = value

& value < 0
--> reaction.lower_bound = value
```

```
direction === "invert"

& value > 0
--> reaction.lower_bound = value

& value < 0
--> reaction.upper_bound = value
``` 

The 'score' parameter is now displayed when toggled on the card.

### DIFF
--- a/src/views/InteractiveMap/InteractiveMap.vue
+++ b/src/views/InteractiveMap/InteractiveMap.vue
@@ -458,14 +458,15 @@ export default Vue.extend({
         }
       });
       // Apply the values from diffFVA results as bounds for simulation
-      const editedBounds = card.manipulations
-        .map(function(manipulation) {
-          // do we need to filter out values that are zero or is this controlled on the backend?
-          const model_reaction =  model.model_serialized.reactions.find(rxn => rxn.id === manipulation.id)
-          const newBound = manipulation.value;
-          switch (manipulation.direction) {
-            case 'up':
-              if (newBound > 0) {
+      const editedBounds = card.manipulations.map(function(manipulation) {
+        // do we need to filter out values that are zero or is this controlled on the backend?
+        const model_reaction = model.model_serialized.reactions.find(
+          rxn => rxn.id === manipulation.id
+        );
+        const newBound = manipulation.value;
+        switch (manipulation.direction) {
+          case "up":
+            if (newBound > 0) {
               return {
                 operation: "modify",
                 type: "reaction",
@@ -475,8 +476,8 @@ export default Vue.extend({
                   upper_bound: model_reaction.upper_bound
                 }
               };
-              };
-              if (newBound < 0) {
+            }
+            if (newBound < 0) {
               return {
                 operation: "modify",
                 type: "reaction",
@@ -486,10 +487,10 @@ export default Vue.extend({
                   lower_bound: model_reaction.lower_bound
                 }
               };
-              };
-              break;
-              case 'down':
-              if (newBound > 0) {
+            }
+            break;
+          case "down":
+            if (newBound > 0) {
               return {
                 operation: "modify",
                 type: "reaction",
@@ -499,8 +500,8 @@ export default Vue.extend({
                   lower_bound: model_reaction.lower_bound
                 }
               };
-              };
-              if (newBound < 0) {
+            }
+            if (newBound < 0) {
               return {
                 operation: "modify",
                 type: "reaction",
@@ -510,10 +511,10 @@ export default Vue.extend({
                   upper_bound: model_reaction.upper_bound
                 }
               };
-              };
-              break;
-              case 'invert':
-              if (newBound > 0) {
+            }
+            break;
+          case "invert":
+            if (newBound > 0) {
               return {
                 operation: "modify",
                 type: "reaction",
@@ -523,8 +524,8 @@ export default Vue.extend({
                   upper_bound: model_reaction.upper_bound
                 }
               };
-              };
-              if (newBound < 0) {
+            }
+            if (newBound < 0) {
               return {
                 operation: "modify",
                 type: "reaction",
@@ -534,17 +535,17 @@ export default Vue.extend({
                   lower_bound: model_reaction.lower_bound
                 }
               };
-              };
-              break;
-          }
-            })
-          // Simulate the model with the updated bounds
-          // adding first all the modifications from the design
-          // and then the modifications (edited bounds) computed above
-          this.postSimulation(card, model, [
-            ...this.cardModifications(card),
-            ...editedBounds
-          ]);
+            }
+            break;
+        }
+      });
+      // Simulate the model with the updated bounds
+      // adding first all the modifications from the design
+      // and then the modifications (edited bounds) computed above
+      this.postSimulation(card, model, [
+        ...this.cardModifications(card),
+        ...editedBounds
+      ]);
     },
     postSimulation(card, model, operations) {
       this.updateCard({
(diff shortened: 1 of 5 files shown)


src/views/InteractiveMap/InteractiveMap.vue
```diff
@@ -457,59 +457,95 @@ export default Vue.extend({
...
-                  card.productionGrowthRate
-              );
+      // Apply the values from diffFVA results as bounds for simulation
+      const editedBounds = card.manipulations.map(function(manipulation) {
+        // do we need to filter out values that are zero or is this controlled on the backend?
+        const model_reaction = model.model_serialized.reactions.find(
+          rxn => rxn.id === manipulation.id
+        );
+        const newBound = manipulation.value;
+        switch (manipulation.direction) {
```
kvikshaug: In general I recommend just using a regular if-elseif statement instead of switch. In short, `switch` can cause subtle bugs when you forget `break`, or when you intentionally leave it out and later modify the code. (Douglas Crockford has a nice section on this in "Javascript: The Good Parts")

src/views/InteractiveMap/InteractiveMap.vue
```diff
@@ -457,59 +457,95 @@ export default Vue.extend({
...
-          // (reference flux * score) / production growth
-          const editedBounds = card.manipulations
-            .map(function(manipulation) {
-              const newBound = Math.round(
-                (wildtypeFluxdistribution[manipulation.id] *
-                  manipulation.value) /
-                  card.productionGrowthRate
-              );
+      // Apply the values from diffFVA results as bounds for simulation
+      const editedBounds = card.manipulations.map(function(manipulation) {
```
kvikshaug: Prefer arrow function here, otherwise `this` won't be available in the scope

```suggestion
      const editedBounds = card.manipulations.map(manipulation => {
```

src/views/InteractiveMap/InteractiveMap.vue
```diff
@@ -457,59 +457,95 @@ export default Vue.extend({
...
+          case "up":
+            if (newBound > 0) {
               return {
                 operation: "modify",
                 type: "reaction",
                 id: manipulation.id,
                 data: {
                   lower_bound: newBound,
-                  upper_bound: newBound
+                  upper_bound: model_reaction.upper_bound
```
kvikshaug: In some cases, I'm getting `model_reaction is undefined` error here:

image

It happens consistently when I visualize a DiffFVA design for one of my jobs, I'd be happy to try to help debug it.

src/views/InteractiveMap/InteractiveMap.vue
ChristianLieven: But I didn't forget the `break` right? Isn't it save then?

src/views/InteractiveMap/InteractiveMap.vue
ChristianLieven: I think it is important that you have to have diffFVA results generated with the newest backend. Did you make sure to test it with recent results?

ChristianLieven:
> I'm still getting zero growth when visualizing the first result in https://caffeine.dd-decaf.eu/jobs/405.

Firefox and Chrome order the results differently. Could you specify exactly which result you tried by naming the parameters/ outcomes?

src/views/InteractiveMap/InteractiveMap.vue
ChristianLieven: Oh but it could also be the added reactions I suppose?

kvikshaug:
I just got a tip from @Midnighter that the cause could be that objective is set to maximizing growth while it should be production. It's the selected job:

image

src/views/InteractiveMap/InteractiveMap.vue
ChristianLieven: Could you confirm that this happened with newer jobs?

kvikshaug:
Now that we don't need to do the wild type simulation, we should set the constraints as manipulations on the card instead of calculating them post-hoc in the `simulateDiffFVACard` function. Then you'd see the correct bounds when hovering reactions in Escher, and maybe there wouldn't even be a need for a separate simulation function for DiffFVA.

src/views/InteractiveMap/InteractiveMap.vue
```diff
@@ -457,59 +457,91 @@ export default Vue.extend({
           hasSimulationError: false
```
lukaw3d: There's unused `const operations = this.cardModifications(card);` above

src/views/InteractiveMap/InteractiveMap.vue
```diff
@@ -457,59 +457,91 @@ export default Vue.extend({
...
+            }
+            if (newBound < 0) {
+              return {
+                operation: "modify",
+                type: "reaction",
+                id: manipulation.id,
+                data: {
+                  upper_bound: newBound,
+                  lower_bound: model_reaction.lower_bound
+                }
```
lukaw3d: Would all those combinations be more readable like this?
```ts
const editedBounds = card.manipulations.map(manipulation => {
  const model_reaction = model.model_serialized.reactions.find(
    rxn => rxn.id === manipulation.id
  );
  const oldLowerBound = model_reaction.lower_bound;
  const oldUpperBound = model_reaction.upper_bound;

  // manipulation.value can never be equal to zero hence we don't need to check for it.
  const newBound = manipulation.value;

  if (
    (manipulation.direction === "up" && newBound > 0) ||
    (manipulation.direction === "invert" && newBound > 0) ||
    (manipulation.direction === "down" && newBound < 0)
  ) {
    return {
      operation: "modify",
      type: "reaction",
      id: manipulation.id,
      data: {
        lower_bound: newBound,
        upper_bound: oldUpperBound
      }
    };
  }

  if (
    (manipulation.direction === "up" && newBound < 0) ||
    (manipulation.direction === "invert" && newBound < 0) ||
    (manipulation.direction === "down" && newBound > 0)
  ) {
    return {
      operation: "modify",
      type: "reaction",
      id: manipulation.id,
      data: {
        lower_bound: oldLowerBound,
        upper_bound: newBound
      }
    };
  }
})

```

ChristianLieven: Certainly would be :O

src/views/InteractiveMap/InteractiveMap.vue
ChristianLieven: Removed!

src/views/InteractiveMap/InteractiveMap.vue
kvikshaug: No it doesn't seem to happen with newer jobs.

kvikshaug: I was able to reproduce this with a newer job after all. It happens when one of the manipulations is for one of the heterologous reactions. You're trying to find that reaction in the wild type model, which of course isn't there. I'll start work on a fix in a separate PR to be merged to this one.

kvikshaug: Fixed in https://github.com/DD-DeCaF/caffeine-vue/pull/117

PR #116, opened 2019-07-18
fix: set objective to the optimization target

annasirunian:
https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1003

### DIFF
--- a/src/views/Jobs/JobResultsTable.vue
+++ b/src/views/Jobs/JobResultsTable.vue
@@ -922,7 +922,9 @@ export default Vue.extend({
               type: cardType,
               // Design card fields
               objective: {
-                reaction: null,
+                reaction: this.prediction.result.reactions[
+                  this.prediction.result.target
+                ],
                 maximize: true
               },
               reactionAdditions: addedReactions,


src/views/Jobs/JobResultsTable.vue
```diff
@@ -922,7 +922,9 @@ export default Vue.extend({
               type: cardType,
               // Design card fields
               objective: {
-                reaction: null,
+                reaction: this.prediction.result.reactions[
+                  this.prediction.result.target
```
lukaw3d: If this can still end up undefined, I would make it explicit, e.g.
```ts
reaction: this.prediction.result.target
  ? this.prediction.result.reactions[this.prediction.result.target]
  : null // or :undefined
```

kvikshaug: - On the backend side it was set to return empty string, not null (but I guess checking for truthiness works for that too)
- `null` for objective means that the objective is growth. So this would hide an error and just give wrong simulation results
- This should never be the case, so I don't think it should be handled - or if it is, an error should be thrown so it's reported to Sentry and we can investigate

lukaw3d: If I understand:
- backend job should fail and log sentry instead of returning `target = ""`
- that way `objective.reaction` always has a value in new jobs
- either remove old jobs (having `target = undefined`), or make their `objective.reaction` more explicitly undefined

kvikshaug: >     * backend job should fail and log sentry instead of returning `target = ""`

Well, it is possible that the backend finds no pathways, and in that case, we don't know the target. So it should be allowed to pass it as empty or null.

But if there are no pathways, the frontend should never try to access `target` ([currently handled here](https://github.com/DD-DeCaF/caffeine-vue/blob/devel/src/views/Jobs/JobResultsTable.vue#L3)). However when there are one or more pathways, *then* `target` should always be defined.

>     * that way `objective.reaction` always has a value in new jobs

:+1: 

>     * either remove old jobs (having `target = undefined`), or make their `objective.reaction` more explicitly undefined

Yes, and one of the cards in the sprint is indeed about removing all old jobs. (I wouldn't make their reaction undefined, but we could probably write a data migration to backpopulate their values - it should be easy to find the target for old jobs too)

PR #117, opened 2019-07-18
Apply DiffFVA modifications correctly

kvikshaug:
- Moves the DiffFVA logic from the simulation step back to being resolved when creating the card, and sets the modifications on the actual card.
- Now there's no need for the `simulateDiffFVA` function in the interactive map component, so reuse the simulation logic for a design card. (We should do the same for data driven cards actually.)
- Fix: When finding the original bounds for the manipulations, search also in the card's added reactions (so that the heterologous reactions can be found).

### DIFF
--- a/src/views/Jobs/JobResultsTable.vue
+++ b/src/views/Jobs/JobResultsTable.vue
@@ -982,6 +982,7 @@ export default Vue.extend({
             } else if (
               jobPrediction.method === "PathwayPredictor+DifferentialFVA"
             ) {
+              // Apply the knockouts
               jobPrediction.knockouts.forEach(reactionId => {
                 this.$store.dispatch("interactiveMap/knockoutReaction", {
                   uuid: card.uuid,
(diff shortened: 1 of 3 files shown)


src/views/Jobs/JobResultsTable.vue
```diff
@@ -1003,11 +1003,32 @@ export default Vue.extend({
                 manipulation => {
                   // First find the original bounds for the reaction, because one of them
                   // will need to be part of the request to modify bounds.
-                  const model_reaction = this.model.model_serialized.reactions.find(
+                  let oldLowerBound;
+                  let oldUpperBound;
+                  // Try to find the reaction among the card's added reactions, in case
```
lukaw3d: Do we need the same in `PathwayPredictor+CofactorSwap` https://github.com/DD-DeCaF/caffeine-vue/blob/2aef6ab47b5f1f5f9bc8b8b2f27ab4ca83c7442b/src/views/Jobs/JobResultsTable.vue#L1069-L1081

kvikshaug: I don't think any heterologous reactions would be used for the swap? /cc @Midnighter

Midnighter: That *can* actually happen. If any heterologous reaction uses NADH or NADPH and it turns out that in the organism's metabolism the other co-factor performs better then it will become a swap target.

kvikshaug: Created card: https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1027

PR #118, opened 2019-07-19
Recreate escher when changing maps to preserve state

kvikshaug:
There have been multiple reported issues of state not being synchronized
when changing maps. Additionally, there was the "double vision" bug. And
finally, [knockouts from visualized design jobs were not always shown
correctly on the map](https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1008).

This change completely destroys and recreates Escher when map changes
and appears to solve all above issues.

As a consequence, the `onEscherReady` promise is no longer relevant.
Instead, watchers will check if the builder exists, and abort if it
doesn't. Then, when Escher is ready, state is always reapplied to make
sure that data for any aborted watcher is applied correctly.

### DIFF
--- a/src/views/InteractiveMap/Escher.vue
+++ b/src/views/InteractiveMap/Escher.vue
@@ -45,6 +45,7 @@ export default Vue.extend({
   props: ["mapData", "card", "isLoadingMap"],
   data: () => ({
     escherBuilder: null,
+    escherReady: false,
     initializingEscher: false,
     hasBoundsError: false,
     defaultReactionStyles: ["color", "size", "text", "abs"],
@@ -175,6 +176,7 @@ export default Vue.extend({
           zoom_extent_canvas: true,
           first_load_callback: () => {
             this.initializingEscher = false;
+            this.escherReady = true;
 
             // Update the current map state.
             // Important: Any new logic to set map state that is added below, needs to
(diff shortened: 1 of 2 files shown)


Midnighter:
Any experience on how this affects performance, for example, switching between maps or playing the changes between cards?

kvikshaug:
> Any experience on how this affects performance, for example, switching between maps or playing the changes between cards?

It adds between 1-2 seconds when changing maps, but the `load_map` call is now included in initializing, and that took about 1 second before. The other processes (downloading the map and re-applying card state) also take some time, so it's not a very noticeable difference.

Playing between cards are not affected, this is only about changing maps.

kvikshaug:
I found there is a race condition where watchers may try to set data on the builder after it is created but before the map is finished loading, I will do some more work on the branch to fix this.

kvikshaug:
To reproduce the knockout bug:

1. Find a job with knockouts
2. Visualize it
3. Change map to one which includes the knocked out reactions

The knockout is correctly applied in the simulation, but not visualized with a red cross in Escher, even though [`set_knockout_reactions`](https://github.com/DD-DeCaF/caffeine-vue/blob/51acd5ab367fc871f420e0e514c1a1a4a028c438/src/views/InteractiveMap/Escher.vue#L271) seems to be called correctly.

kvikshaug:
Dropping this in favor of https://github.com/DD-DeCaF/escher/pull/27

PR #121, opened 2019-07-24
Show snackbar message for developers when hotreloading

lukaw3d:


Midnighter:
`module.hot` only exists/is only defined in development?

lukaw3d:
Yes, webpack-dev-server

src/App.vue
```diff
@@ -255,7 +259,8 @@ export default Vue.extend({
     drawer: false,
     isProjectCreationDialogVisible: false,
     isOrganismCreationDialogVisible: false,
-    isExpanded: true
+    isExpanded: true,
+    hotreloadStatus: undefined
```
kvikshaug: I don't know if it makes much of a difference but at least in the past we tried to prefer `null` over `undefined` consistently where possible.

lukaw3d: Oh, right :+1:

PR #123, opened 2019-07-24
Let the user know if the model is infeasible

kvikshaug:
Depends on https://github.com/DD-DeCaF/simulations/pull/140

### DIFF
--- a/src/store/modules/interactiveMap.ts
+++ b/src/store/modules/interactiveMap.ts
@@ -57,6 +57,7 @@ export interface Card {
   // General simulation fields
   isSimulating: boolean;
   hasSimulationError: boolean;
+  solverStatus: string | null;
   growthRate: number | null;
   fluxes: number | null;
   // When adding a card from the interactive map, the dialog should be opened
--- a/src/views/InteractiveMap/Card.vue
+++ b/src/views/InteractiveMap/Card.vue
@@ -36,6 +36,19 @@
       </v-btn>
     </v-toolbar>
     <v-card-title primary-title class="py-2" v-if="isSelected">
+      <v-alert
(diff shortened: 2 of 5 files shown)


src/store/modules/interactiveMap.ts
```diff
@@ -57,6 +57,7 @@ export interface Card {
   // General simulation fields
   isSimulating: boolean;
   hasSimulationError: boolean;
+  solverStatus: string | null;
```
Midnighter: `hasSimulationError` could then just be a computed of `!!solverStatus && solverStatus !== "optimal"`? Maybe not worth all the refactoring.

src/views/InteractiveMap/Card.vue
```diff
@@ -36,6 +36,14 @@
       </v-btn>
     </v-toolbar>
     <v-card-title primary-title class="py-2" v-if="isSelected">
+      <v-alert
+        :value="card.solverStatus && card.solverStatus !== 'optimal'"
```
Midnighter: Could this use `hasSimulationError`?

src/views/InteractiveMap/Card.vue
```diff
@@ -259,6 +267,11 @@ export default Vue.extend({
       if (this.isSelected) {
         if (this.card.hasSimulationError) {
           return "error";
+        } else if (
+          this.card.solverStatus &&
```
Midnighter: Could this use `hasSimulationError`?

src/views/InteractiveMap/Card.vue
kvikshaug: Not really, that's used for handling HTTP-level errors like network issues or unhandled backend exceptions.

PR #124, opened 2019-07-24
Speedup escher by using 3d transformations for panning

lukaw3d:


Midnighter:
Does it have a noticeable effect?

lukaw3d:
Not on my pc :D

PR #126, opened 2019-07-25
feat: add export feature

annasirunian:


src/views/Jobs/JobResultsTable.vue
```diff
@@ -1254,6 +1262,39 @@ export default Vue.extend({
...
+          link.setAttribute("download", `${fileName}.csv`);
+          document.body.appendChild(link);
+          link.click();
+        });
+      });
+    },
+    getFormattedTime() {
+      const today = new Date();
+      const year = today.getFullYear();
+      // JavaScript months are 0-based.
```
lukaw3d: :laughing:

src/views/Jobs/JobResultsTable.vue
```diff
@@ -1254,6 +1262,39 @@ export default Vue.extend({
...
+      const month = today.getMonth() + 1;
+      const day = today.getDate();
+      const hour = today.getHours();
+      const minute = today.getMinutes();
+      return year + "-" + month + "-" + day + "-" + hour + "-" + minute;
```
lukaw3d: Luckily, there's `this.$moment().format('YYYY-MM-DD-HH-mm')`

src/views/Jobs/JobResultsTable.vue
```diff
@@ -1254,6 +1262,39 @@ export default Vue.extend({
...
+        }).then(response => {
+          // Force browser to download a file
+          const url = window.URL.createObjectURL(new Blob([response.data]));
+          const link = document.createElement("a");
+          link.href = url;
+          const fileName =
+            jobPrediction.method.split("+")[1] + "-" + this.getFormattedTime();
+          link.setAttribute("download", `${fileName}.csv`);
+          document.body.appendChild(link);
+          link.click();
```
lukaw3d: Remove link after :)

src/views/Jobs/JobResultsTable.vue
```diff
@@ -1254,6 +1262,39 @@ export default Vue.extend({
...
+      this.selected.map(jobPrediction => {
+        axios({
+          url: `${settings.apis.metabolicNinja}/predictions/${
+            this.prediction.id
+          }/${jobPrediction.id}`,
+          method: "GET",
+          responseType: "blob"
+        }).then(response => {
+          // Force browser to download a file
+          const url = window.URL.createObjectURL(new Blob([response.data]));
```
lukaw3d: Would this work?
(so javascript doesn't have to do any downloading, decoding, re-encoding work. Significant for large files)
```ts
      this.selected.forEach(jobPrediction => {
        const url = `${settings.apis.metabolicNinja}/predictions/${
          this.prediction.id
        }/${jobPrediction.id}`;

        const link = document.createElement("a");
        link.href = url;
        link.target = "_blank";

        // set filename on backend via `response.headers["Content-Disposition"] = "attachment; filename=\"filename.csv\""`

        document.body.appendChild(link);
        link.click();
        document.body.removeChild(link);
      });
```

lukaw3d: Scratch that- Chrome might block them as behaving as multiple spam popups. This would work for a single zip file, but then you can just have `<v-btn :href="url">Export</v-btn>`

kvikshaug: > then you can just have `<v-btn :href="url">Export</v-btn>`

Indeed. Since we will have a single zip file, I don't see a need to pre-download in memory using ajax, or to create fake anchor elements. Just build a url dynamically like that with IDs as GET parameters or similar.

annasirunian: Discussed with Luka, the thing is we can't send authorization header using that approach, see here: https://stackoverflow.com/questions/23485857/download-anchor-link-with-authorization-header

Decided to stick with the current approach, unless you object

src/views/Jobs/JobResultsTable.vue
```diff
@@ -1254,6 +1262,32 @@ export default Vue.extend({
...
       }
       return swaps;
+    },
+    exportData() {
+      const predictionIds = this.selected.map(
+        jobPrediction => jobPrediction.id
+      );
+      axios({
+        url: `${settings.apis.metabolicNinja}/predictions/export`,
+        method: "POST",
```
kvikshaug: This should be a `GET` request when just fetching data

src/views/Jobs/JobResultsTable.vue
kvikshaug: Ugh, that's true. There are other workarounds, but then I think you're right to choose this approach. Can you then please add a comment before the `axios` call to explain in detail why we chose this solution?

PR #128, opened 2019-07-25
Hide objective and flux on DiffFVA cards showing score

lukaw3d:
Related to https://github.com/DD-DeCaF/caffeine-vue/pull/125

### DIFF
--- a/src/views/InteractiveMap/Card.vue
+++ b/src/views/InteractiveMap/Card.vue
@@ -69,12 +69,8 @@
         </v-layout>
       </v-container>
 
-      <!-- Shared between Design cards and DiffFVA -->
-      <v-container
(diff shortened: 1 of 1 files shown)


src/views/InteractiveMap/Card.vue
```diff
@@ -189,11 +177,11 @@
                 slot="activator"
               >
                 <template v-slot:label>
-                  <div v-if="!showDiffFVAScore">
-                    Show manipulation targets
+                  <div v-if="showDiffFVAScore">
```
kvikshaug: Just a suggestion: you could also use a single element with a binding here, `:style="{ color: showDiffFVAScore ? 'black' : null }"`

PR #133, opened 2019-07-30
Install google analytics

kvikshaug:
For development, I chose to not load vue-analytics at all, instead of loading it with debug features ([google docs](https://developers.google.com/analytics/devguides/collection/analyticsjs/debugging) and [vue-analytics docs](https://matteogabriele.gitbooks.io/vue-analytics/content/docs/debug.html)). I felt it was spamming the console a bit much. I'm open to re-enabling it if you think it's more useful than noisy though. My thinking is that it's not useful right now since we don't really send any custom events, we just rely on the default implementation tracking navigation and that's it.

image


lukaw3d:
Make it opt-in for GDPR?
https://medium.com/dailyjs/google-analytics-gdpr-and-vuejs-e1bd6affd2b4

kvikshaug:
> Make it opt-in for GDPR?
> https://medium.com/dailyjs/google-analytics-gdpr-and-vuejs-e1bd6affd2b4

Made a call on this with Niko that we'll merge now without dealing with GDPR, but add a card for later to:

- Extend the privacy policy
- Anonymize personal user data (like IP address)
- Add cookie notice with opt-in

PR #135, opened 2019-08-09
Reset password

annasirunian:
https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1036

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -233,6 +233,14 @@
       <v-snackbar v-model="hotreloadStatus" :timeout="6000" top right>
         Hotreload status: {{ hotreloadStatus }}
       </v-snackbar>
+
+      <v-snackbar
+        color="success"
+        v-model="isPasswordResetSuccess"
+        :timeout="9000"
+      >
+        Your password has been successfully changed.
+      </v-snackbar>
     </v-app>
   </div>
 </template>
@@ -305,6 +313,14 @@ export default Vue.extend({
       set() {
         // Ignore passed value argument.
         this.$store.commit("setUnauthorizedError", null);
+      }
+    },
+    isPasswordResetSuccess: {
+      get() {
+        return this.$store.state.passwordResetSuccess;
+      },
+      set() {
+        this.$store.commit("setPasswordResetSuccess");
       }
     },
     availableProjects(): ColoredProjectItem[] {
--- a/src/App.vue
+++ b/src/App.vue
@@ -320,7 +320,7 @@ export default Vue.extend({
         return this.$store.state.passwordResetSuccess;
       },
       set() {
-        this.$store.commit("setPasswordResetSuccess");
+        this.$store.commit("setPasswordResetSuccess", false);
       }
     },
     availableProjects(): ColoredProjectItem[] {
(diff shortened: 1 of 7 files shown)


src/App.vue
```diff
@@ -307,6 +315,14 @@ export default Vue.extend({
         this.$store.commit("setUnauthorizedError", null);
       }
     },
+    isPasswordResetSuccess: {
+      get() {
+        return this.$store.state.passwordResetSuccess;
+      },
+      set() {
+        this.$store.commit("setPasswordResetSuccess");
```
lukaw3d: This is too strange. v-snackbar will attempt to set `isPasswordResetSuccess = false`, which will in turn commit passwordResetSuccess=true to store.

Does the notification even ever disappear?

annasirunian: Yes, the notification disappears in 9 seconds, but you are right, the logic was strange, it fixed now

lukaw3d: Just because this was strange to me, I looked into snackbar implementation. It has an internal variable `isActive` that mirrors v-model (your isPasswordResetSuccess)
(~ `watch(v-model) { isActive = value }`).

So that's why notification still disappeared, even though isPasswordResetSuccess=true when it tried to set it to false after 9 seconds.

So sideeffect of `v-snackbar will attempt to set isPasswordResetSuccess = false, which will in turn commit passwordResetSuccess=true to store` isn't that notification doesn't disappear (rescued by isActive), but that it will never appear again (doomed by watch)

:confused: That's a tougher bug to spot

src/views/ResetRequest.vue
```diff
@@ -0,0 +1,99 @@
...
+    requestResetLink() {
+      this.isRequestingResetLink = true;
+      axios
+        .post(`${settings.apis.iam}/reset_request`, { email: this.email.value })
+        .then((response: AxiosResponse) => {
+          this.resetRequestSuccessMessage = response.data;
+          this.isResetRequestSuccess = true;
+        })
+        .catch(error => {
+          this.resetRequestErrorMessage = error.response.data;
```
kvikshaug: This assumes that the network request succeeded (if it doesn't, `error.response` isn't set, see [the axios docs](https://github.com/axios/axios#handling-errors)), just add some general error message if response isn't set. Here's a suggestion: "We were unable to send the email. Please check your internet connection, or try again in a few minutes."

src/views/PasswordReset.vue
```diff
@@ -0,0 +1,150 @@
...
+        if (!v) {
+          return "Confirm your password";
+        }
+        return true;
+      }
+    }
+  }),
+  created() {
+    this.token = this.$route.params.token;
+    // Verify token validity
```
kvikshaug: Ah, I see now what this endpoint was used for. The expiry is readable by the client though so we don't need to make a request to check it

src/views/PasswordReset.vue
```diff
@@ -0,0 +1,150 @@
...
+    },
+    changePassword() {
+      this.isRequestingPasswordChange = true;
+      axios
+        .post(`${settings.apis.iam}/password_reset/${this.token}`, {
+          password: this.password.value
+        })
+        .then((response: AxiosResponse) => {
+          this.$router.push({ name: "home" });
+          this.$store.commit("setPasswordResetSuccess", true);
```
kvikshaug: It could be convenient for the user to be automatically logged in here and not have to log in again.

PR #138, opened 2019-08-13
Make sorting more consistent across browsers

lukaw3d:


kvikshaug:
Hmm, it doesn't seem right to have to implement basic algorithms like sorting directly in a high level project like this. Maybe consider adding underscorejs or a similar third party lib instead?

lukaw3d:
I prefer that too
Superseeded by https://github.com/DD-DeCaF/caffeine-vue/pull/139

PR #144, opened 2019-09-02
Use vue-rx to fix response order and add debounce in MNX reaction autocomplete

lukaw3d:
https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1014

I'm not too happy about adding the complexity of Rx (and increasing package size), but combining promise ordering, debouncing, mutating isLoading, and requestError seems hard to write expressively otherwise

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11777,6 +11777,14 @@
         "rx-lite": "*"
       }
     },
+    "rxjs": {
+      "version": "6.5.2",
+      "resolved": "https://registry.npmjs.org/rxjs/-/rxjs-6.5.2.tgz",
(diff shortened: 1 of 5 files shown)


ChristianLieven:
I am frankly wholly inexperienced with `rxjs` and don't think I have the time to invest in learning it to an appreciable level within the next three month at the very least. On the other hand, I doubt I'll get around to contributing to the software development process for about the same amount of time. Therefore, I am mostly impartial but leaning towards 'let's not introduce a new paradigm that the majority of us may not be intuitively familiar with'.

annasirunian:
I think that if rxjs is the right tool to use in such situation, then I'm fine with that, though I also don't feel confident with the technology yet. But from my side I'm ready to maintain and contribute to the part of code if it's written with observables

lukaw3d:
Superseeded by https://github.com/DD-DeCaF/caffeine-vue/pull/152

PR #152, opened 2019-09-20
MNX search debouncing & ordering

kvikshaug:
Here's my suggestion of how to achieve debouncing and order guarantee without RxJS.

- For debouncing, turns out it's straight forward to use [`lodash`](https://lodash.com/docs/4.17.15#debounce) which is already a dependency.
- To guarantee search results are for the latest query, a unique ID is assigned to the request promise and compared when it resolves. If the ID has changed the results are ignored in favor of the results of another request.


kvikshaug:
Agreed with @annasirunian to merge this to staging now in order to have it ready for the user review

PR #153, opened 2019-09-20
fix: call `finally`, not `then`, for settled promise handler

kvikshaug:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally


lukaw3d:
+1 I was looking at those too, but worried about browser support

kvikshaug:
Ah right. We need IE11 compatibility and [it's not even in edge](https://developer.microsoft.com/en-us/microsoft-edge/platform/status/promiseprototypefinally/) so I was a bit too quick here

PR #156, opened 2019-09-25
Submit reaction and compound ids that were found in the models

annasirunian:
https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1059

### DIFF
--- a/src/components/AutocompleteMnxMetabolite.vue
+++ b/src/components/AutocompleteMnxMetabolite.vue
@@ -44,6 +44,8 @@ export interface MetaNetXMetabolite {
   formula: string;
   modelNames?: Array<string>;
   displayValue?: string;
+  foundId?: string; // exists if was found in the passed models
+  namespace?: string;
 }
 
 export interface Annotation {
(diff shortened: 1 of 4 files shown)


src/components/AutocompleteMnxReaction.vue
```diff
@@ -185,12 +199,15 @@ export default Vue.extend({
               )
             )
             .then((response: any) => {
```
lukaw3d: This definitely needs caching - it is repeatedly requesting full models in each autocomplete

You can probably use `withFullModel`

src/components/AutocompleteMnxMetabolite.vue
```diff
@@ -119,29 +129,33 @@ export default Vue.extend({
...
           response.data.forEach((metabolite: MetaNetXMetabolite) => {
-            const annotation = metabolite.annotation;
-            const metaboliteIds = [metabolite.mnx_id];
-            for (const namespace in annotation) {
-              annotation[namespace].forEach(metaboliteId =>
-                metaboliteIds.push(metaboliteId)
-              );
-            }
+            const deprecatedIds = metabolite.annotation["deprecated"] || [];
+            delete metabolite.annotation["deprecated"];
```
lukaw3d: I would only delete it out of metaboliteIdsMap / reactionIdsMap

src/components/AutocompleteMnxMetabolite.vue
```diff
@@ -119,29 +129,33 @@ export default Vue.extend({
...
-              ) {
-                isMetaboliteFound = true;
-                metabolite.modelNames.push(modelName);
+            for (const model in this.metabolitesInModelsMap) {
+              const [modelId, modelName] = JSON.parse(model);
+              for (const namespace in metaboliteIdsMap) {
+                metaboliteIdsMap[namespace].forEach(metaboliteId => {
+                  if (this.metabolitesInModelsMap[model].has(metaboliteId)) {
+                    isMetaboliteFound = true;
+                    metabolite.modelNames.push(modelName);
```
lukaw3d: This appends the same modelName multiple times if multiple ids from annotations are found

src/components/AutocompleteMnxReaction.vue
```diff
@@ -248,7 +265,7 @@ export default Vue.extend({
       }
       this.selectedValue = selectedReaction;
       const reaction: Reaction = {
-        id: selectedReaction.reaction.mnx_id,
+        id: selectedReaction.foundId || selectedReaction.reaction.mnx_id,
```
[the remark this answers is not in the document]
annasirunian: `mnx_id` will still be passed into `Add a reaction...`, since we don't pass `modelIds` prop from parent component

PR #159, opened 2019-09-30
Delete row

annasirunian:
https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1052

New package: vuetify-confirm which allows to get confirmation from user when deleting condition or sample with related data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "vue-moment": "^4.0.0",
     "vue-router": "^3.0.6",
     "vuetify": "^1.5.14",
+    "vuetify-confirm": "^2.0.0",
     "vuex": "^3.1.1"
   },
   "devDependencies": {
--- a/src/components/NewExperiment.vue
+++ b/src/components/NewExperiment.vue
@@ -917,7 +917,7 @@ export default Vue.extend({
         sample =>
           sample.condition && sample.condition.temporaryId === conditionId
       );
-      let isConfirmationRequired = relatedSamples.length ? true : false;
+      const isConfirmationRequired = relatedSamples.length > 0;
 
       const confirmation = !isConfirmationRequired
         ? Promise.resolve(true)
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -24,6 +24,7 @@
     "vue-moment": "^4.0.0",
     "vue-router": "^3.0.6",
     "vuetify": "^1.5.14",
+    "vuetify-confirm": "^2.0.0",
```
lukaw3d: @kvikshaug I think their abstraction of dialogs as a promise are worth it. We talked with Anna that we could copy a subset of their code and use it for input dialogs too

kvikshaug: Well, I'd be on board with the idea of forking the whole thing - we'll still get the abstraction and avoid the risks.

src/components/NewExperiment.vue
```diff
@@ -238,7 +238,7 @@
                       </td>
                       <td>
                         <v-text-field
-                          v-model.number="fluxomicsItem.measurement"
```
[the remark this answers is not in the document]
annasirunian: There is a bug in `v-model.number`. When trying to enter negative decimal number between 0 and 1, it converts `-0` to `0`

src/components/NewExperiment.vue
```diff
@@ -758,11 +877,112 @@ export default Vue.extend({
...
         temporaryId: uuidv4()
       });
     },
+    deleteCondition(index) {
+      const conditionId = this.tables.conditions.items[index].temporaryId;
+      if (
+        this.tables.samples.items.some(
+          sample =>
+            sample.condition && sample.condition.temporaryId === conditionId
+        )
```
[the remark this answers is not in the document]
annasirunian: Thanks for a great suggestion!

package.json
annasirunian: Agreed with Luka to keep the package for now and replace it with our custom component later

PR #160, opened 2019-10-01
Autoselect exact match

annasirunian:
Pasted reactions and compounds which have the exact match with the result from metanetx service (mnx and all ids from the annotation) are now autoselected.

TODO:
Highlight autocompletes without matches

### DIFF
--- a/src/components/AutocompleteMnxMetabolite.vue
+++ b/src/components/AutocompleteMnxMetabolite.vue
@@ -36,7 +36,7 @@
 <script lang="ts">
 import Vue from "vue";
 import axios from "axios";
-import { debounce } from "lodash";
+import { debounce, flatten } from "lodash";
 import uuidv4 from "uuid/v4";
 import { Prop } from "vue/types/options";
 import * as settings from "@/utils/settings";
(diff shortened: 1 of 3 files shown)


src/components/AutocompleteMnxReaction.vue
```diff
@@ -167,7 +211,7 @@ export default Vue.extend({
             if (isReactionFound) {
               searchResultsInTheModel.push(reaction);
             } else {
-              reaction.namespace = "metanetx.chemical";
+              reaction.namespace = "metanetx.reaction";
```
lukaw3d: Good catch

src/components/AutocompleteMnxMetabolite.vue
```diff
@@ -178,6 +224,22 @@ export default Vue.extend({
...
+            const firstResult = this.searchResults[0];
+            const metaboliteIds = new Set(
+              Object.values(firstResult.annotation).flat()
+            );
+            metaboliteIds.add(firstResult.mnx_id);
+            if (
+              metaboliteIds.has(this.searchQuery) ||
+              this.searchQuery === firstResult.name
+            ) {
+              this.selectedValue = firstResult;
```
lukaw3d: Not needed?

src/components/AutocompleteMnxMetabolite.vue
```diff
@@ -178,6 +224,22 @@ export default Vue.extend({
             );
           }
           this.searchResults.push(...searchResultsNotInTheModel);
+          // If pasted metabolite has the exact match with the first result
+          // from metanetx service, it should be autoselected
+          if (this.searchQuery === this.forceSearchQuery) {
+            const firstResult = this.searchResults[0];
+            const metaboliteIds = new Set(
+              Object.values(firstResult.annotation).flat()
```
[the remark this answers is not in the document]
kvikshaug: Ah, good catch. Let's use the lodash version to ensure compatibility for IE users

src/components/AutocompleteMnxMetabolite.vue
annasirunian: 👍

PR #162, opened 2019-10-03
Empty autocompletes

annasirunian:
Done:
1. Fix `required` rule for custom components
2. Mark autocompletes that have no matches between pasted data and results from metanetx service

Todo:
1. Ask user which condition or sample pasting data belongs to

### DIFF
--- a/src/components/AutocompleteMnxMetabolite.vue
+++ b/src/components/AutocompleteMnxMetabolite.vue
@@ -229,25 +229,27 @@ export default Vue.extend({
           // If pasted metabolite has the exact match with the first result
           // from metanetx service, it should be autoselected
           if (this.searchQuery === this.forceSearchQuery) {
-            [searchResultsNotInTheModel[0], searchResultsInTheModel[0]].forEach(
-              searchResult => {
-                if (!searchResult) {
-                  return;
-                }
-                const metaboliteIds = new Set(
-                  flatten(Object.values(searchResult.annotation))
-                );
-                metaboliteIds.add(searchResult.mnx_id);
-                if (
-                  metaboliteIds.has(this.searchQuery) ||
-                  this.searchQuery === searchResult.name
-                ) {
-                  this.searchResults = [searchResult];
-                  this.selectedItem = searchResult;
-                  this.onChange(this.selectedItem);
-                }
+            const exactMatch = [
+              searchResultsInTheModel[0],
+              searchResultsNotInTheModel[0]
+            ].find(searchResult => {
+              if (!searchResult) {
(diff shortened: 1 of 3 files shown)


src/components/NewExperiment.vue
```diff
@@ -26,7 +26,7 @@
                   <v-data-table
                     :headers="selectedTable.headers"
                     :items="tables.conditions.items"
-                    :pagination.sync="pagination"
```
lukaw3d: I would keep something like:
```
:pagination.sync="disabledToPreventWrongIndexOnSecondPage"

// We are relying on data table's `index` (for currentRowIndex and @paste)
// but that only works correctly on the first page.
disabledToPreventWrongIndexOnSecondPage: undefined
```
to avoid future bugs

src/components/NewExperiment.vue
```diff
@@ -855,6 +860,9 @@ export default Vue.extend({
       this.selectedTable.items.push({
         temporaryId: uuidv4()
       });
+      setTimeout(() => {
+        this.$refs.newExperimentForm.validate();
+      }, 0);
```
lukaw3d: Note to self: implement this nicely by extending v-form with e.g.
```
<v-form validate-before-focused>
  watch
    inputs
        this.validate()
```

src/components/AutocompleteMnxMetabolite.vue
```diff
@@ -230,7 +229,7 @@ export default Vue.extend({
           // If pasted metabolite has the exact match with the first result
           // from metanetx service, it should be autoselected
           if (this.searchQuery === this.forceSearchQuery) {
-            [searchResultsInTheModel[0], searchResultsNotInTheModel[0]].forEach(
+            [searchResultsNotInTheModel[0], searchResultsInTheModel[0]].forEach(
```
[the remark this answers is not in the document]
annasirunian: 👍

src/components/AutocompleteMnxMetabolite.vue
```diff
@@ -229,25 +229,28 @@ export default Vue.extend({
...
+              );
+              const exactValues = new Set(metaboliteIds);
+              exactValues.add(searchResult.mnx_id);
+              exactValues.add(searchResult.name);
+
+              return exactValues.has(this.searchQuery);
+            });
+
+            if (exactMatch) {
+              this.searchResults.push(exactMatch);
```
lukaw3d: I think it is already in there :D

annasirunian: True, will remove this

PR #163, opened 2019-10-07
feat: replace vuetify-confirm with custom component

annasirunian:
https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1068

### DIFF
--- a/src/components/ConfirmDialog.vue
+++ b/src/components/ConfirmDialog.vue
@@ -1,0 +1,53 @@
+<template>
+  <v-dialog @input="change" value="true" max-width="500px">
+    <v-card class="text-xs-center pa-3">
+      <v-icon large color="warning" class="warning-icon">warning</v-icon>
+      <v-card-text class="headline">{{ message }}</v-card-text>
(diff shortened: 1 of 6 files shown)


src/components/ConfirmDialog.vue
```diff
@@ -0,0 +1,53 @@
+<template>
+  <v-dialog @input="change" value="true" max-width="500px">
+    <v-card class="text-xs-center pa-3">
+      <v-icon large color="warning" class="warning-icon">warning</v-icon>
+      <v-card-text class="headline">{{ message }}</v-card-text>
+      <v-card-actions>
+        <v-spacer></v-spacer>
+        <v-btn color="primary" flat="flat" @click="choose(false)">
```
kvikshaug: The false response could maybe be a different color

annasirunian: Agree, but that's the way we usually have Cancel button (following Material Design)

PR #171, opened 2019-10-16
map ecModel fluxes

kvikshaug:
Closes https://github.com/DD-DeCaF/scrum/issues/1128

### DIFF
--- a/src/views/InteractiveMap/Escher.vue
+++ b/src/views/InteractiveMap/Escher.vue
@@ -27,7 +27,7 @@
         equal to the lower bound.
       </v-snackbar>
     </div>
-    <Legend :card="card" />
+    <Legend :card="card" :ecModel="model && model.ec_model" />
   </div>
 </template>
 
@@ -250,6 +250,13 @@ export default Vue.extend({
       } else {
         this.escherBuilder.load_model(this.model.model_serialized);
       }
+
+      // We cannot highlight missing reactions for ecModels, since most of the
+      // identifiers on the map would not match those in the model.
+      this.escherBuilder.settings.set(
+        "highlight_missing",
+        !this.model.ec_model
+      );
     },
     setReactionAdditions() {
       if (!this.card) {
--- a/src/views/InteractiveMap/Escher.vue
+++ b/src/views/InteractiveMap/Escher.vue
@@ -374,17 +374,19 @@ export default Vue.extend({
           // For single enzymes, use the flux in the reaction "XXXNo1" (for
           // this, check first that no arm reaction is already present).
           const rootRxn = rxn.substring(0, rxn.length - 3);
-          if (!(rootRxn in fluxesMapped)) {
+          if (!("arm_" + rootRxn in fluxes)) {
             newRxn = rootRxn;
           }
         }
         fluxesMapped[newRxn] = fluxes[rxn];
-        if (newRxn.endsWith("_REV") && fluxes[rxn] > 0) {
+      });
+      Object.keys(fluxesMapped).forEach(rxn => {
+        if (rxn.endsWith("_REV") && fluxesMapped[rxn] > 0) {
           // For reversible enzymes ("XXX_REV") that are active in the backwards
           // direction, replace the existing flux in the forward reaction by the
           // negative value.
-          newRxn = newRxn.substring(0, newRxn.length - 4);
-          fluxesMapped[newRxn] = -fluxes[rxn];
+          const forwardRxn = rxn.substring(0, rxn.length - 4);
+          fluxesMapped[forwardRxn] = -fluxesMapped[rxn];
         }
       });
       return fluxesMapped;
--- a/src/views/InteractiveMap/Escher.vue
+++ b/src/views/InteractiveMap/Escher.vue
@@ -369,11 +369,11 @@ export default Vue.extend({
         let newRxn = rxn;
         if (rxn.startsWith("arm_")) {
           // For isozymes, use the flux in the reaction "arm_XXX".
-          newRxn = rxn.substring(4);
+          newRxn = rxn.slice(4);
         } else if (rxn.endsWith("No1")) {
           // For single enzymes, use the flux in the reaction "XXXNo1" (for
           // this, check first that no arm reaction is already present).
-          const rootRxn = rxn.substring(0, rxn.length - 3);
+          const rootRxn = rxn.slice(0, -3);
           if (!("arm_" + rootRxn in fluxes)) {
             newRxn = rootRxn;
           }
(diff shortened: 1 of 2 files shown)


BenjaSanchez:
@kvikshaug done on my side! I added the mapping logic and works well locally. I've done a more thorough testing of the [python mapper](https://nbviewer.jupyter.org/github/BenjaSanchez/notebooks/blob/master/ec_model_mapper_tests.ipynb) I had created before, but I checked manually and all cases are preserved on this mapper (single enzyme, isozymes, reversible without enzymes, reversible with enzymes).

src/views/InteractiveMap/Escher.vue
```diff
@@ -349,6 +360,37 @@ export default Vue.extend({
...
+          }
+        }
+        fluxesMapped[newRxn] = newFlux;
+        if (newRxn.endsWith("_REV") && fluxes[rxn] > 0) {
+          // For reversible enzymes ("XXX_REV") that are active in the backwards
+          // direction, replace the existing flux in the forward reaction by the
+          // negative value.
+          newRxn = newRxn.substring(0, newRxn.length - 4);
+          newFlux = newFlux * -1;
+          fluxesMapped[newRxn] = newFlux;
```
kvikshaug: I think `Object.keys()` order is not deterministic, so could this be overwritten by the forwards direction with the same name if it comes after the reverse?

src/views/InteractiveMap/Escher.vue
```diff
@@ -349,6 +360,37 @@ export default Vue.extend({
...
+        var newRxn = rxn;
+        var newFlux = fluxes[rxn];
+        if (rxn.startsWith("arm_")) {
+          // For isozymes, use the flux in the reaction "arm_XXX".
+          newRxn = rxn.substring(4);
+        } else if (rxn.endsWith("No1")) {
+          // For single enzymes, use the flux in the reaction "XXXNo1" (for
+          // this, check first that no arm reaction is already present).
+          const rootRxn = rxn.substring(0, rxn.length - 3);
+          if (!Object.keys(fluxesMapped).includes(rootRxn)) {
```
lukaw3d:
```suggestion
          if (!(rootRxn in fluxesMapped)) {
```

src/views/InteractiveMap/Escher.vue
```diff
@@ -349,6 +360,37 @@ export default Vue.extend({
...
       return fluxesFiltered;
     },
+    ecModelFluxes(fluxes) {
+      // Map fluxes from an enzyme-constrained model to the corresponding
+      // escher map, as reaction ids from ecModels have additional prefixes
+      // and/or suffixes.
+      const fluxesMapped = {};
+      Object.keys(fluxes).forEach(rxn => {
+        var newRxn = rxn;
+        var newFlux = fluxes[rxn];
```
kvikshaug: Prefer to limit scope by using `const` for immutable vars and `let` for mutable ones. (`newFlux` could be immutable if you assign `newFlux * -1` directly on line 389)

src/views/InteractiveMap/Escher.vue
lukaw3d: If it is already present, you keep `XXXNo1: flux` as is?

src/views/InteractiveMap/Escher.vue
```diff
@@ -349,6 +360,37 @@ export default Vue.extend({
...
+        if (newRxn.endsWith("_REV") && fluxes[rxn] > 0) {
+          // For reversible enzymes ("XXX_REV") that are active in the backwards
+          // direction, replace the existing flux in the forward reaction by the
+          // negative value.
+          newRxn = newRxn.substring(0, newRxn.length - 4);
+          newFlux = newFlux * -1;
+          fluxesMapped[newRxn] = newFlux;
+        }
+      });
+      return fluxesMapped;
```
lukaw3d: Might be a bug:
- `arm_XXX_REV: 1` splits into `XXX_REV: 1, XXX: -1`
- `XXXNo1_REV: 1` splits into `XXXNo1_REV: 1, XXXNo1: -1` (expected `XXX_REV: 1, XXX: -1`)

src/views/InteractiveMap/Escher.vue
lukaw3d: Instead of changing the variable, I'd just use `fluxesMapped[newRxn] = -fluxes[rxn];`

lukaw3d: Oh, and should it be `fluxesMapped[newRxn] = -fluxes[newRxn]`?

src/views/InteractiveMap/Escher.vue
BenjaSanchez: yes, because it would mean that `arm_XXX` was already added as `XXX`

src/views/InteractiveMap/Escher.vue
BenjaSanchez: no, `fluxes[newRxn]` in that case would be = 0, that's why we are updating with the backwards value

src/views/InteractiveMap/Escher.vue
lukaw3d: Okay

If you want code looking more like python version of same code:
```py
#Single enzyme:
elif rxn_id.endswith("No1") and rxn_id[:-3] not in parsed_fluxes.index:
    new_id = rxn_id[:-3]
```

```ts
} else if (rxn.endsWith("No1") && !(rxn.slice(0, -3) in fluxesMapped)) {
  // For single enzymes, use the flux in the reaction "XXXNo1" (for
  // this, check first that no arm reaction is already present).
  newRxn = rxn.slice(0, -3)
}
```

src/views/InteractiveMap/Escher.vue
BenjaSanchez: fixed in [`3027223`](https://github.com/DD-DeCaF/caffeine-vue/pull/171/commits/30272236089b7fdba081fff39e5ccf59f2f2ff70): by doing the replacement of isozymes and reversibilities in 2 separate `forEach` blocks the order in which the reactions come will not matter. Also, by matching `arm_XXX` in the old dict instead of `XXX` in the new one ([line 377](https://github.com/DD-DeCaF/caffeine-vue/pull/171/files#diff-cb5e31d152203986c650570c10bfa708R377)), the order of isozymes and single enzymes will also not matter.

src/views/InteractiveMap/Escher.vue
BenjaSanchez: done in [11f367e](https://github.com/DD-DeCaF/caffeine-vue/pull/171/commits/11f367e0de167fc02d655e776b9dac260cd8b432)

src/views/InteractiveMap/Escher.vue
BenjaSanchez: > Instead of changing the variable, I'd just use `fluxesMapped[newRxn] = -fluxes[rxn];`

done in [11f367e](https://github.com/DD-DeCaF/caffeine-vue/pull/171/commits/11f367e0de167fc02d655e776b9dac260cd8b432)

src/views/InteractiveMap/Escher.vue
BenjaSanchez: > If you want code looking more like python version of same code:

done in [`e439c4b`](https://github.com/DD-DeCaF/caffeine-vue/pull/171/commits/e439c4bcd966bf86533764282d8b2e3fae41c20c)

PR #172, opened 2019-10-16
Validation

annasirunian:


src/components/NewExperiment.vue
```diff
@@ -738,13 +773,27 @@
...
+                  <v-layout>
+                    <v-radio
+                      :label="table.name"
+                      :value="key"
+                      color="primary"
+                      class="mb-1"
+                    ></v-radio>
+                    <v-icon
+                      v-if="
+                        isTableValid[table.value] &&
```
lukaw3d: You can replace `isTableValid[table.value]` with `isTableValid[key]`

or even move isTableValid into tables[*].isValid and use `table.isValid`

annasirunian: 👍

src/components/NewExperiment.vue
```diff
@@ -850,7 +850,8 @@ function getInitialState() {
       name: null,
       description: null,
       project_id: null,
-      id: null
+      id: null,
+      isValid: true
```
lukaw3d: I think this is passed into experiment payload

PR #178, opened 2019-10-22
feat: display organism name in strains autocomplete

annasirunian:
Closes https://github.com/DD-DeCaF/scrum/issues/1138


kvikshaug:
@annasirunian there is a race condition here if the organisms are not yet loaded (can happen if you open the /experiments page directly), you might need to wait for the `organismsPromise`.

image

annasirunian:
@kvikshaug, thanks for the catch!

PR #179, opened 2019-10-22
Order upload

annasirunian:
Closes https://github.com/DD-DeCaF/scrum/issues/1142

### DIFF
--- a/src/components/NewExperiment.vue
+++ b/src/components/NewExperiment.vue
@@ -767,7 +767,7 @@
               </v-autocomplete-extended>
               <v-radio-group v-model="selectedTableKey">
                 <div v-for="(table, key) in tables" :key="key">
-                  <v-tooltip bottom :disabled="!isTableDisabled(key)">
+                  <v-tooltip bottom :disabled="Boolean(!isTableDisabled(key))">
                     <template v-slot:activator="{ on }">
                       <v-layout v-on="on">
                         <v-radio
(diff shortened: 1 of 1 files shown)


src/components/NewExperiment.vue
```diff
@@ -1489,6 +1495,34 @@ export default Vue.extend({
...
+          return false;
+        }
+      }
+      return true;
+    },
+    getMessageforDisabledTable(key) {
+      if (key === "samples") {
+        return "You should enter valid condition first";
+      }
+      return "You should enter valid sample first";
```
lukaw3d: It is usually better to merge these messages with logic that determines isTableDisabled, e.g.
```ts
    isTableDisabled(key): false | string {
      if (key === "conditions") {
        return false;
      } else if (key === "samples") {
        const mainField = this.tables.conditions.mainField;
        if (
          this.tables.conditions.items[0][mainField] &&
          this.tables.conditions.isValid
        ) {
          return false;
        } else {
          return "You should enter valid condition first";
        }
      } else {
        const mainField = this.tables.samples.mainField;
        if (
          this.tables.samples.items[0][mainField] &&
          this.tables.samples.isValid
        ) {
          return false;
        } else {
          return "You should enter valid sample first";
        }
      }
    },
```

or, in more complex situations, `return { disabled: true, msg: "because.." }`

annasirunian: Perfect, thanks!

PR #180, opened 2019-10-23
fix: reset highlight_missing when model is missing

lukaw3d:


lukaw3d:
Not really, this just avoid an error `Cannot set property 'ec_model' of undefined`, 
and keeps a consistent state when there's no model (it doesn't depend on what last model contained)

PR #182, opened 2019-10-24
Fix infinite requests in mnx autocomplete (v3)

lukaw3d:
v1: https://github.com/DD-DeCaF/caffeine-vue/pull/160 fixed infinite requests when input was not focused (paste two rows)
v2: https://github.com/DD-DeCaF/caffeine-vue/pull/173 fixed infinite requests when input lost focus (click outside before loaded)
v3: fixes infinite requests when input is null (click outside after loaded)

@kvikshaug First two commits are for you :D

### DIFF
--- a/src/components/AutocompleteMnxMetabolite.vue
+++ b/src/components/AutocompleteMnxMetabolite.vue
@@ -144,67 +144,92 @@ export default Vue.extend({
       // results from a stale request.
       const searchId = uuidv4();
       this.activeSearchID = searchId;
-      this.isLoading = false;
-      this.requestError = false;
-      this.skipVuetifyClearSearch = false;
 
-      if (this.searchQuery === null || this.searchQuery.trim().length === 0) {
-        this.searchResults = [];
-        return;
-      }
-      if (
-        this.selectedItem &&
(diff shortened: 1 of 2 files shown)


lukaw3d:
:smile: 

Yes, because Promise.resolve().then() helps so that anything returned inside `.then(() => { ... })` is normalized to a promise, and then second section can contain shared logic
```ts
.then(() => {
  if (searchQuery === displayValue) {
    return [this.selectedItem]; // synchronous
  }
  return axios.get().then(() => results) // asynchronous
})
.then(value => {
  // handle both cases with same logic
})
.catch()
.then(() => {
  this.isLoading = false;
  this.skipVuetifyClearSearch = false;
})
```

otherwise you usually have duplicate code for synchronous results
```ts
  if (searchQuery === displayValue) {
    this.isLoading = false;
    this.skipVuetifyClearSearch = false;
    return [this.selectedItem];
  }

  return axios.get().then(() => promisedResults)
    .catch()
    .then(() => {
      this.isLoading = false;
      this.skipVuetifyClearSearch = false;
    })
```

PR #183, opened 2019-10-24
Proteomics

kvikshaug:
- Add proteomics table in new experiment dialog
  - Uses new Uniprot component which matches against https://www.uniprot.org/uniprot/ (service doesn't provide search, so needs to be exact match)
- Show proteomics in data driven card dialog
- Unrelated fix: Better display if strain has no genotype modifications

https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1082
Related backend change: https://github.com/DD-DeCaF/warehouse/pull/20

### DIFF
--- a/src/components/UniprotInput.vue
+++ b/src/components/UniprotInput.vue
@@ -2,12 +2,12 @@
   <div>
     <v-text-field
       label="Protein"
+      v-model="searchQuery"
       :loading="isLoading"
       :hint="hint()"
       persistent-hint
       :append-outer-icon="protein ? 'info' : ''"
       :rules="[requestErrorRule(requestError), ...(rules || [])]"
-      @input="inputChanged"
       @click:append-outer="dialog = true"
       @paste="$emit('paste', $event)"
     ></v-text-field>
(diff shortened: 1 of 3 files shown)


BenjaSanchez:
[add proteomics](https://trello.com/c/8IyIC5Fu/3-add-proteomics)

annasirunian:
Copy/paste doesn't work for `Protein` field, for `Reaction` and `Compound` fields in other tables we used `forceSearchQuery` input property to pass `_pastedText` into`AutocompleteMnxReaction`/`AutocompleteMnxMetabolite` components and make search request to the backend

src/components/UniprotInput.vue
```diff
@@ -86,6 +88,44 @@ export default Vue.extend({
...
+    },
+    searchQuery: {
+      // Watcher needs to be immediate to trigger when copy-paste creates
+      // new rows with forceSearchQuery already set
+      immediate: true,
+      handler() {
+        // Reset state
+        this.protein = null;
+        this.requestError = false;
+        this.$emit("change", this.protein);
```
annasirunian: I would emit `change` event only if exact match was found, then in `NewExperiment` component, `protein` field would still have property `_pastedText` and error message would be `No match was found` instead of `Required`

kvikshaug: That would be better, but then how do we signal that the previous match object was removed, when the user is typing manually? If I remove this event, if I first type a valid query, and then change it, then there is no message:

image

lukaw3d: Maybe if you emit null in `if (!this.searchQuery)` and on `404`

lukaw3d: *not on 404

lukaw3d: No, wait, nevermind

PR #185, opened 2019-10-25
feat: add validation for datetime fields

annasirunian:
Closes https://github.com/DD-DeCaF/scrum/issues/1131


lukaw3d:
For cleaner code, you could turn it into a component
- only pass format once "YYYY-MM-DD HH:mm" (and toLowerCase for placeholder)
- checks validity inside
- returns Date object into the v-model (then no need for 
  `end_time: sample.endTime ? this.$moment(sample.endTime, "YYYY-MM-DD HH:mm").toDate() : null`

PR #187, opened 2019-11-01
Small features

annasirunian:
Done:
- Add input clear functionality to reaction, compound and protein fields
- Replace dialog with tooltip for protein field


src/components/AutocompleteMnxMetabolite.vue
```diff
@@ -363,6 +368,9 @@ export default Vue.extend({
       // takes away focus and vuetify tries to clear search.
       this.skipVuetifyClearSearch = true;
       this.searchQuery = this.forceSearchQuery;
+    },
+    onClear() {
+      this.$emit("clear");
```
kvikshaug: Could also just inline this

PR #190, opened 2019-11-07
Bulk requests for proteomics data

annasirunian:
Part of https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1152.

Done: refactor components to support batch retrieval of protein data from UniProt service.

For now `v-model` for `UniprotInput` isn't implemented, so pasted data only saved in the `NewExperiment` component's data, but isn't displayed.

Updated: pasted proteins are now displayed

### DIFF
--- a/src/components/NewExperiment.vue
+++ b/src/components/NewExperiment.vue
@@ -606,6 +606,7 @@
                                 proteomicsItem
                               )
                             ]"
+                            :passedProtein="proteomicsItem.protein"
                           />
                         </td>
                         <td>
@@ -1357,7 +1358,9 @@ function getInitialState() {
                     identifier: parsedResponse[str]["Entry name"] || "Unknown",
                     name: parsedResponse[str]["Protein names"] || "Unknown",
                     gene:
-                      parsedResponse[str]["Gene names  (primary )"] || "Unknown"
+                      parsedResponse[str]["Gene names  (primary )"] ||
+                      "Unknown",
+                    uniprotId: str
                   };
                 }
                 return { _pastedText: str };
--- a/src/components/NewExperiment.vue
+++ b/src/components/NewExperiment.vue
@@ -606,10 +606,6 @@
                                 proteomicsItem
                               )
                             ]"
-                            :forceSearchQuery="
-                              proteomicsItem.protein &&
-                                proteomicsItem.protein._pastedText
-                            "
                           />
                         </td>
                         <td>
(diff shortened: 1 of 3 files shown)


src/components/NewExperiment.vue
```diff
@@ -1333,9 +1329,43 @@ function getInitialState() {
           { value: "actions", width: "5%" }
         ],
         parsePasted: {
-          protein: str => ({ _pastedText: str }),
-          measurement: str => parseFloat(str),
-          uncertainty: str => parseFloat(str)
+          protein: strs => {
+            const bodyFormData = new FormData();
+            bodyFormData.set("uploadQuery", strs.join(" "));
```
lukaw3d: `FormData.append` has much much better browser support than `FormData.set`

src/components/NewExperiment.vue
```diff
@@ -1699,14 +1702,40 @@ export default Vue.extend({
...
-        // parsedRows = [[["name", "a"], ["measurement", 5], ["uncertainty", null]]]
-        parsedRows.forEach((rowPairs, rowIx) => {
-          if (!table.items[rowOffset + rowIx]) {
-            // Create excess rows.
-            table.items.push({ temporaryId: uuidv4() });
-          }
-          [...rowPairs, ...rowPairsFromDialog].forEach(([property, value]) => {
-            Vue.set(table.items[rowOffset + rowIx], property, value);
+        // Transpose the array, so we can make batch requests
+        const columns = zip(...rows);
```
lukaw3d: This might fail for tens of thousands of rows (max number of arguments given to a function).
Try `unzip(rows)` instead

src/components/NewExperiment.vue
```diff
@@ -1699,14 +1702,40 @@ export default Vue.extend({
...
+        // TODO: handle excess columns
+        const parsedColumnsPromises = columns.map((column, columnIx) => {
+          const property = table.headers[columnOffset + columnIx].value;
+          const parsedColumnPromise = table.parsePasted[property](column, {
```
lukaw3d: Oh, we can't have `if (cell) { parse } else { null }` now that multiple cells (=column) is parsed. We don't handle pasting empty cells well anymore.

And I'm not a fan of delegating it into every parsePasted function
```diff
-strs => strs.map(str => parseFloat(str))
+strs => strs.map(str => str ? parseFloat(str) : null)
```

TODO
- [ ] think

src/components/NewExperiment.vue
```diff
@@ -606,10 +606,6 @@
                                 proteomicsItem
                               )
                             ]"
-                            :forceSearchQuery="
-                              proteomicsItem.protein &&
-                                proteomicsItem.protein._pastedText
```
lukaw3d: (I guess this might get replaced with v-model (or reintroduced) for pasting mismatches)

src/components/UniprotInput.vue
```diff
@@ -45,7 +47,22 @@ export default Vue.extend({
   computed: {},
   watch: {
     searchQuery() {
-      this.triggerQuery();
+      if (!this.skipQueryTrigger) {
```
lukaw3d: How about `if (this.searchQuery !== this.protein.uniprotId) {`

annasirunian: Will work with a little change:
```
searchQuery() {
      if (
        this.passedProtein &&
        this.searchQuery === this.passedProtein.uniprotId
      ) {
        return;
      }
      this.triggerQuery();
    },
```

PR #193, opened 2019-11-14
Bulk requests

annasirunian:
Closes https://github.com/DD-DeCaF/scrum/issues/1152


src/components/AutocompleteMnxReaction.vue
```diff
@@ -388,48 +315,10 @@ export default Vue.extend({
...
-            compartment: "",
-            stoichiometry: m.coefficient
-          };
-        }),
-        namespace: selectedReaction.namespace
-      };
-
-      this.$emit("change", {
-        reaction: reaction,
-        mnxReaction: selectedReaction
```
lukaw3d: Issue: we are using it in `CardDialogDesign/addReaction` method

src/components/NewExperiment.vue
```diff
@@ -1302,11 +1304,34 @@ function getInitialState() {
...
         ],
         parsePasted: {
-          // Temporarily create mock reaction object and use forceSearchQuery;
-          // selecting a reaction then clears forceSearchQuery.
-          reaction: str => ({ _pastedText: str }),
-          measurement: str => parseFloat(str),
-          uncertainty: str => parseFloat(str)
+          reaction: (strs, { reactionsInModelsMap }) =>
+            axios
+              .get(`${settings.apis.metanetx}/reactions/batch?query=${strs}`)
```
lukaw3d: Formalize array to string conversion with `strs.join(',')`

src/components/NewExperiment.vue
```diff
@@ -1717,7 +1787,32 @@ export default Vue.extend({
...
+                this.reactionsInModelsMap[key].add(reaction.id)
+              );
+              this.metabolitesInModelsMap[key] = new Set([]);
+              model.model_serialized.metabolites.forEach(metabolite =>
+                this.metabolitesInModelsMap[key].add(
+                  getMetaboliteId(metabolite.id, metabolite.compartment)
+                )
+              );
+            });
+            return selected ? [["sample", selected]] : [];
```
lukaw3d: If selected can be false-y, above `selected.condition.strain.organism_id` will fail

src/components/NewExperiment.vue
```diff
@@ -1717,7 +1787,32 @@ export default Vue.extend({
...
+            )
+          ).then(() => {
+            modelIds.forEach(modelId => {
+              const model = this.getModelById(modelId);
+              const key = JSON.stringify([model.id, model.name]);
+              this.reactionsInModelsMap[key] = new Set([]);
+              model.model_serialized.reactions.forEach(reaction =>
+                this.reactionsInModelsMap[key].add(reaction.id)
+              );
+              this.metabolitesInModelsMap[key] = new Set([]);
```
lukaw3d: Should probably reset `this.metabolitesInModelsMap = {}` at some point

src/components/NewExperiment.vue
lukaw3d: Add comment that we force user to choose a `sample`, so that `parsePastedMetabolites(strs)` only has to consider model ids for that sample, instead of handling different model ids because each row could have a different sample.

src/components/AutocompleteMnxReaction.vue
annasirunian: Thanks for the catch!

src/components/NewExperiment.vue
annasirunian: `selected` cannot be falsy now when there is a specific order to upload data.
The main reason we force user to choose a `sample` is because then we can map metanetx ids to ones that are in the model, otherwise simulation will fail

lukaw3d: Oh, yeah, I was thinking that it was for scenario:
- already have multiple rows with differing samples (=> differing model ids for mapping)
- paste columns over those rows
- so force single sample to make mapping mnx ids easy

But actual scenario is:
- paste new rows
- if no selected sample, can't map mnx ids
- select sample later, mnx ids aren't re-mapped

PR #195, opened 2019-11-21
Data driven dialog

kvikshaug:
It looks like a big changeset, but most of the changes are in the second commit which simply moves the `CardDialogDataDriven` out of the parent component, `CardDialog`. Now it's reimplementing logic from the parent, which makes it easier to control when simulations happens etc.

[The issue](https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/912) is implemented in [the third commit](https://github.com/DD-DeCaF/caffeine-vue/pull/195/commits/cbc13dd4bd24e0f96f0c10e15d4789cc0e625e74) by removing the organism drop-down and now auto-selecting the organism based on the selected strain.

The last commit implements the status [I posted about here](https://biosustain.slack.com/archives/G8SBED3C3/p1574331459000700).

### DIFF
--- a/src/views/InteractiveMap/CardDialogDataDriven.vue
+++ b/src/views/InteractiveMap/CardDialogDataDriven.vue
@@ -10,40 +10,13 @@
             <v-flex xs6>
               <p class="text-xs-right subheading mb-0 mt-2 font-italic">
                 Simulation status:
-                <span v-if="card.solverStatus">
-                  <span v-if="card.solverStatus == 'optimal'">Successful</span>
-                  <span v-else-if="card.solverStatus == 'infeasible'"
-                    >Infeasible</span
-                  >
-                </span>
-                <span v-else-if="card.sampleErrors.length > 0"
-                  >Can not integrate data</span
-                >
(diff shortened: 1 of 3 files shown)


src/views/InteractiveMap/CardDialogDataDriven.vue
```diff
@@ -3,7 +3,57 @@
...
+              <p class="text-xs-right subheading mb-0 mt-2 font-italic">
+                Simulation status:
+                <span v-if="card.solverStatus">
+                  <span v-if="card.solverStatus == 'optimal'">Successful</span>
+                  <span v-else-if="card.solverStatus == 'infeasible'"
+                    >Infeasible</span
+                  >
+                </span>
+                <span v-else-if="card.sampleErrors.length > 0"
+                  >Can not integrate data</span
```
lukaw3d: (moving these many conditions into e.g. getStatusMessage method (or computed) would be easier to read (+ `<v-progress-circular v-if="modfying||simulating">`))

PR #196, opened 2019-11-21
Validate all pages

annasirunian:
Closes https://github.com/DD-DeCaF/scrum/issues/1153

We are defining `rules` property on each table.
Based on that we are checking if all items in particular table are valid.

### DIFF
--- a/src/components/NewExperiment.vue
+++ b/src/components/NewExperiment.vue
@@ -1072,11 +1072,11 @@
                           :disabled="!!whyIsTableDisabled(key)"
                         ></v-radio>
                         <v-icon
-                          v-if="table.isValid() && !isTableUnchanged(table)"
+                          v-if="isTableValid(table) && !isTableUnchanged(table)"
                           color="success"
                           >done</v-icon
                         >
-                        <v-icon v-if="!table.isValid()" color="error"
+                        <v-icon v-if="!isTableValid(table)" color="error"
                           >error_outline</v-icon
                         >
                       </v-layout>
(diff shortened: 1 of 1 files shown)


src/components/NewExperiment.vue
```diff
@@ -1595,6 +1714,10 @@ export default Vue.extend({
         .then(response => {
           const mnxMetabolites: MetaNetXMetabolite[] = response.data;
           return mnxMetabolites.map((mnxMetabolite, index) => {
+            // Handle pasting empty cells
+            if (strs[index] === "") {
+              return null;
+            }
```
lukaw3d: This may be unreliable

Check that both batch mnx and uniprot results don't get misaligned.
`query=4,,1,3` and `uploadQuery=4  1 3` would have to return `[{id: 4}, anything, {id: 1}, {id: 3}]`, but I'd expect `[{id: 4}, {id: 1}, {id: 3}]`

src/components/NewExperiment.vue
```diff
@@ -1473,9 +1568,29 @@ function getInitialState() {
...
         items: [{ temporaryId: uuidv4() }],
         isValid() {
-          return true;
+          const rules = this.rules;
+          const keys = Object.keys(rules);
+          return this.items.every(item =>
+            keys.every(key =>
+              rules[key](item).every(valid => typeof valid === "boolean")
+            )
+          );
```
lukaw3d: I guess try to reuse the same function, or remove isValid from tables and make methods: isTableValid(table)

src/components/NewExperiment.vue
annasirunian: Metanetx and Uniprot services behave differently:
1. mnx - returns empty object if no match was found (it also checks EC number and when we are sending empty string it can return any object which has EC equals to `''`). We are iterating through response data safely when mapping and handling pasting empty cells
2. uniprot - doesn't return anything if no match was found. That's why we are iterating through initial query

kvikshaug: > it also checks EC number and when we are sending empty string it can return any object which has EC equals to `''`

Should we change that to disallow search with empty string? For example by returning HTTP 400, or just leave it `null`?

annasirunian: I think it's better with `null`, so pasting empty cells will clear the fields as we discussed

PR #198, opened 2019-11-22
Show invalid rows first

annasirunian:
Closes https://github.com/DD-DeCaF/scrum/issues/1154

Proposed solution: to add button that sorts array to display invalid rows first.

Issues with initial idea to add a checkbox to filter the array:
1. If checkbox is selected, the moment user changes the row to be valid, that row will disappear, that might be confusing. Keeping that row displayed is also not a perfect UX solution considering that checkbox "Show only invalid rows" is selected
2. Possibility to paste data will be disabled

Update:
Pagination was changed, now user can navigate to first or last page.

### DIFF
--- a/src/components/NewExperiment.vue
+++ b/src/components/NewExperiment.vue
@@ -1028,6 +1028,9 @@
               class="mt-3"
               >Add row</v-btn
             >
+            <v-btn color="primary" small @click="showInvalidRows()" class="mt-3"
+              >Show invalid rows first</v-btn
+            >
           </v-card>
         </v-flex>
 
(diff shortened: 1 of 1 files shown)


src/components/NewExperiment.vue
```diff
@@ -2172,6 +2142,30 @@ export default Vue.extend({
     },
     onMeasurementClear(index, table, property) {
       table.items[index][property] = null;
+    },
+    showInvalidRows() {
+      const table = this.tables[this.selectedTableKey];
+      const keys = Object.keys(table.rules);
+      const validItems: any[] = [];
+      const invalidItems: any[] = [];
```
kvikshaug: Unused

PR #202, opened 2019-11-28
Reaction glow to visualize enzyme usage

kvikshaug:
https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1160

Some necessary refactoring included adjusting bounds in the frontend's copy of the model based on operations, so that we have the correct bounds set by the adapter. They are then later used to calculate the enzyme usages.

Slider looks like this and defaults to 90%:

image

There is an open question whether performance is good enough - the application hangs for a few seconds after simulations complete on my machine, and adjusting the slider is not very responsive, but it works.

@BenjaSanchez: The `enzymeUsageMapped` property maps the ecModel identifiers back to the ones used in bigg, based on how we do it with the fluxes. Can you confirm whether the implementation is correct?


src/views/InteractiveMap/Escher.vue
```diff
@@ -139,6 +164,81 @@ export default Vue.extend({
...
+      Object.keys(this.enzymeUsage).forEach(id => {
+        let newId = id;
+        if (id.startsWith("arm_")) {
+          newId = id.slice(4);
+        } else if (id.endsWith("No1")) {
+          newId = id.slice(0, -3);
+        }
+        mappedReactions[newId] = this.enzymeUsage[id];
+      });
+      return mappedReactions;
```
lukaw3d: (or)
```ts
import {mapKeys} from "lodash";
return mapKeys(this.enzymeUsage, id => id.replace(/^arm_/, "").replace(/No1$/, ""));
```

src/views/InteractiveMap/Escher.vue
```diff
@@ -139,6 +164,81 @@ export default Vue.extend({
...
+    enzymeUsageMapped() {
+      // Map the ecModel-specific reaction identifiers back to the original
+      // identifiers, so that they are correctly recognized in the Escher map.
+      const mappedReactions = {};
+      Object.keys(this.enzymeUsage).forEach(id => {
+        let newId = id;
+        if (id.startsWith("arm_")) {
+          newId = id.slice(4);
+        } else if (id.endsWith("No1")) {
+          newId = id.slice(0, -3);
```
BenjaSanchez: @kvikshaug note that a `No1` reaction should only be chosen if there is no associated `arm` reaction in the whole network (as it is done in `ecModelFluxes()`), otherwise you are over-writing usage of the `arm` reaction with the value in `No1` (which might be 0 if `No2`, `No3`, etc. are chosen by the simulations instead). So the function should be something like:

```suggestion
          const rootId = id.slice(0, -3);
          if (!("arm_" + rootId in this.enzymeUsage)) {
            newId = rootId;
          }
```

Note that compared to `ecModelFluxes()` we can disregard the reversible reactions (`_REV`), as those will just contain redundant information in terms of enzyme usage (the GPR rule of the backward and forward reaction are equivalent).

kvikshaug:
Thanks for your comments! It should be addressed in the most recent commits.

> Do we need another special effect in escher, alongside highlight?

For reference, [results from yesterday's planning meeting discussion](https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1088)

> So I will do further testing once we are able to upload proteomics+growth rate.

Sounds good. I'll wait with merging this PR and mark the issue as blocked until we can investigate those odd cases.

src/views/InteractiveMap/Card.vue
```diff
@@ -228,6 +228,21 @@
         </v-layout>
       </v-container>
 
+      <!-- ecModels visualizing enzyme usage -->
+      <v-container v-if="showEnzymeUsageSlider" fluid class="pa-0">
+        <v-layout row>
+          <v-flex>
+            Highlight reactions where enzyme usage exceeds the following
+            threshold:
```
BenjaSanchez:
```
Highlight reactions where enzyme usage is greater than or equal to the following threshold:
```

src/views/InteractiveMap/Escher.vue
```diff
@@ -302,6 +417,24 @@ export default Vue.extend({
...
+      // NOTE: This currently intereferes with fluxomics since we use
+      // highlighted reactions for both use cases, so currently the assumption
+      // is that users won't upload both fluxomics and proteomics in the same
+      // experiment.
+      if (!this.enzymeUsage) {
+        this.escherBuilder.set_highlight_reactions([]);
+        return;
+      }
+
+      // Only highlight reactions with enzyme usage above the given threshold.
```
BenjaSanchez:
```
// Only highlight reactions with enzyme usage greater than or equal to the given threshold.
```

src/views/InteractiveMap/Escher.vue
```diff
@@ -302,6 +417,24 @@ export default Vue.extend({
         this.card.sample.fluxomics.map(m => m.reaction_identifier)
       );
     },
+    setEnzymeUsage() {
+      // Based on proteomics and simulated fluxes, visualizes enzyme usage by
+      // highlighting reactions over the given threshold.
```
BenjaSanchez:
```
// highlighting reactions with enzyme usage greater than or equal to the given threshold.
```

src/views/InteractiveMap/Legend.vue
```diff
@@ -51,12 +51,18 @@
...
+        <v-layout align-center v-show="!showScoreLegend && !showEnzymeUsage">
           <div class="measured-flux"></div>
           <p class="mb-0 ml-3">
             Measured flux
           </p>
         </v-layout>
+        <v-layout align-center v-show="showEnzymeUsage">
+          <div class="measured-flux"></div>
+          <p class="mb-0 ml-3">
+            Enzyme usage exceeds threshold
```
BenjaSanchez:
```
Reaction with enzyme usage greater than or equal to threshold
```
Hopefully the legend still looks decent with the longer description?

kvikshaug: image

WDYT? It looks a bit long to me but maybe I'm just too used to the previous size. At least this makes it easier to distinguish the different gradient colors.

kvikshaug:
Summary of the latest commits:

* Updated documentation and legend to reflect the intended threshold logic
* Fixed issue where a reaction's usage value would not reflect the highest of its genes
* Individual genes are now also highlighted
* Gene highlighting is reflected in the legend
* Related Escher PR https://github.com/DD-DeCaF/escher/pull/30

Note: There is not an escher release yet so the branch's escher version is out of date. To test it, a custom escher build must be made. Here's a screenshot of what it currently looks like:

image

src/views/InteractiveMap/Legend.vue
BenjaSanchez: LGTM!

kvikshaug:
> Looks great! I tested for genes/reactions at different levels, comparing to enzyme usages predicted directly from the simulations functions in [this notebook](https://github.com/BenjaSanchez/notebooks/blob/e732e7de1055d4aa1f7272071f3075fbdeddf1e1/ec_model_simulations_tests/notebook.ipynb) (section 3) and saw consistent results.

:tada: 

> Also, no perceivable lag in my machine when visualizing gene glow :)

I did find and fix a glitch that caused unnecessary recalculations, so that's good to get confirmed!

> Slightly unrelated: ATM we don't have any way of knowing the actual enzyme usage value for each gene, we can only infer it by playing around with the slider and seeing what happens with the glow. I know that each gene has a "data" feature (see below), how hard would it be to adapt that to show the usage percentage in any scenario where we do gecko simulations? This ofc could be addressed in a separate PR.

Escher has a [`set_gene_data`](https://escher.readthedocs.io/en/latest/javascript_api.html#escher.Builder.set_gene_data) function, but when called, it uses the values for the fluxes. So the actual fluxes are ignored, and all reactions change color according to the enzyme usage numbers. As far as I can tell, there doesn't seem to be an easy way currently to achieve what we want.

We could try a different approach though, like trying to show the percentage together with the gene name.

BenjaSanchez:
> We could try a different approach though, like trying to show the percentage together with the gene name

sounds good! Let's wait for the review meeting for further discussion, but in the meantime this PR can be merged :)

BenjaSanchez:
@kvikshaug ps we already have a card to address visualizing usage values: https://github.com/DD-DeCaF/scrum/issues/1089

PR #204, opened 2019-12-20
Proteomics data visualization

annasirunian:
Closes https://github.com/DD-DeCaF/scrum/issues/583.

Note: can be tested when https://github.com/DD-DeCaF/warehouse/pull/23 is merged.


src/views/InteractiveMap/Card.vue
```diff
@@ -294,6 +326,7 @@ export default Vue.extend({
     showMethodHelpDialog: false,
     isCardDialogVisible: false,
     showDiffFVAScore: false,
+    showProteomicsData: false,
```
kvikshaug: No need to duplicate this field, just use a computed property with getter and setter which directly accesses the value in the store.

src/components/NewExperiment.vue
```diff
@@ -1673,7 +1673,24 @@ export default Vue.extend({
...
+                  : [],
+                orderedLocus: parsedResponse[str][
+                  "Gene names  (ordered locus )"
+                ]
+                  ? parsedResponse[str]["Gene names  (ordered locus )"].split(
+                      " "
+                    )
+                  : [],
+                orf: parsedResponse[str]["Gene names  (ORF )"]
+                  ? parsedResponse[str]["Gene names  (ORF )"].split(" ")
```
kvikshaug: What is the deal with these weird column keys from uniprot :man_facepalming:

PR #208, opened 2020-01-13
Ensure exact id match on uniprot response results

kvikshaug:
Note, I was able to confirm that this is not necessary for the batch requests.

### DIFF
--- a/src/components/UniprotInput.vue
+++ b/src/components/UniprotInput.vue
@@ -113,7 +113,7 @@ export default Vue.extend({
           // multiple results. We care about exact identifier matches, so try to
           // find one among the results.
           const parsedResponse = tsvParse(response.data).find(
-            item => item["Entry"] === uniprotId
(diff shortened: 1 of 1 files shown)


src/components/UniprotInput.vue
```diff
@@ -109,7 +109,17 @@ export default Vue.extend({
           }
         })
         .then(response => {
-          const parsedResponse = tsvParse(response.data)[0];
+          // The API query appears to match loosely on free text and yields
+          // multiple results. We care about exact identifier matches, so try to
+          // find one among the results.
+          const parsedResponse = tsvParse(response.data).find(
+            item => item["Entry"] === uniprotId
```
BenjaSanchez: [UniProtKB entries only use upper cases](https://www.uniprot.org/help/accession_numbers), so for increased usability you could allow case insensitive matches

PR #211, opened 2020-01-21
Added an Icon with a tooltip as growth rate alert to the Card component

blablabananarama:
Closes https://github.com/DD-DeCaF/scrum/issues/1086

The value chosen from which it starts showing up is an arbitrary low value, maybe it should be more clearly defined? 

Here's how it looks in action: 
grafik

Can whoever reads this also give me some feedback if I worked on this correctly, the way that the team would normally do this?

### DIFF
--- a/src/views/InteractiveMap/Card.vue
+++ b/src/views/InteractiveMap/Card.vue
@@ -181,20 +181,6 @@
       <v-container fluid class="pa-0">
         <v-layout wrap>
           <v-flex>Growth rate:</v-flex>
-          <v-flex>
-            <div v-if="card.growthRate !== null">
-              <v-tooltip v-if="card.growthRate <= 0.00001" bottom>
-                <template v-slot:activator="{ on }">
-                  <v-icon color="warning" dark v-on="on">error</v-icon>
-                </template>
-                <span>
-                  You are simulating with proteomics data but without setting
-                  any growth rate. <br />Consider adding a growth rate to your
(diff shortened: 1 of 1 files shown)


src/views/InteractiveMap/Card.vue
```diff
@@ -181,6 +181,20 @@
       <v-container fluid class="pa-0">
         <v-layout wrap>
           <v-flex>Growth rate:</v-flex>
+          <v-flex>
+            <div v-if="card.growthRate !== null">
+              <v-tooltip v-if="card.growthRate <= 0.00001" bottom>
```
BenjaSanchez: The warning should trigger whenever there are proteomics data + no growth rate provided, regardless of the growth rate value. That way we don't have to decide on a threshold. Sorry if this was not clear.

BenjaSanchez:
Another thing: Note that experimental data can only be integrated in models in **data-driven** cards, so this warning should ONLY show up in those cards. In the example that you provided above it's implemented for a design card, so please change accordingly.

src/views/InteractiveMap/Card.vue
blablabananarama: It should now only trigger whenever no growth rate is provided, and only for data-driven cards with proteomics data provided. Also I moved the icon to the right.

ChristianLieven:
It seems like the way to do semantic commit messages is still not clear to you. Let's meet up so I can explain.

PR #214, opened 2020-01-23
fix: enzyme usage slider only shown for ec models

blablabananarama:
Closes https://github.com/DD-DeCaF/scrum/issues/1206

From a clean branch this time, should hopefully now be able to merge into devel.

### DIFF
--- a/src/views/InteractiveMap/Card.vue
+++ b/src/views/InteractiveMap/Card.vue
@@ -261,11 +261,7 @@
       </v-container>
 
       <!-- ecModels visualizing enzyme usage -->
-      <v-container
-        v-if="showEnzymeUsageSlider && model && model.ec_model"
-        fluid
-        class="pa-0"
-      >
+      <v-container v-if="showEnzymeUsageSlider" fluid class="pa-0">
         <v-layout row>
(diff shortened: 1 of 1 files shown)


src/views/InteractiveMap/Card.vue
```diff
@@ -261,7 +261,11 @@
       </v-container>
 
       <!-- ecModels visualizing enzyme usage -->
-      <v-container v-if="showEnzymeUsageSlider" fluid class="pa-0">
+      <v-container
+        v-if="showEnzymeUsageSlider && model && model.ec_model"
```
BenjaSanchez: I think it makes more sense to include this logic where `showEnzymeUsageSlider` is defined (L521 in `Card.vue`):

```js
    showEnzymeUsageSlider() {
      return (
        this.card.type == "DataDriven" &&
        this.card.sample &&
        this.card.sample.proteomics.length > 0
      );
    },
```

ChristianLieven:
Same comments here, you're not fully grasping the semantic commit messages ;) It's supposed to be present tense so it reads like a changelog when you look at all the commits later.

PR #220, opened 2020-01-29
fix: update privacy policy to include GDPR rights

ChristianLieven:
I auto-generated a new privacy policy from TermsFeed.com but then removed all commercial statements that didn't fit our platform as it currently operates. I've attached the full policy for future reference.

I've also added a GDPR data rights statement and, at the very end, I've added Niko's email address as a point of contact.


ChristianLieven:
Originally generated privacy policy:
[termsfeed-html-privacy-policy.zip](https://github.com/DD-DeCaF/caffeine-vue/files/4129061/termsfeed-html-privacy-policy.zip)

PR #221, opened 2020-01-30
Feat/consent

JuroOravec:


ChristianLieven:
PS: Next time let's avoid multi-branch confusion and progressively add small improvements (rather than developing two/three features that require the same logic in parallel)

JuroOravec:
Snackbar added! Agree with the flowchard

PR #227, opened 2020-02-03
Refactor/consent review feedback

JuroOravec:
- Makes consents logic and components (cookie banner & cookies options in privacy policy) toggable via env var VUE_APP_ENABLE_CONSENTS
- Updates copy in cookie banner and cookies description to state that a user has to agree to preferences cookies to not be shown the cookie banner again
  - Note: currently, only unregistered users have to do so. Registered users will be shown the cookie banner only if the do not have any cookie consents stored in the db. We assume that if they have any cookie consent in the db (accepted or rejected), they must have seen or interacted with the cookies settings before)
- Refactors consents logic out of  "session" store module into a separate "consents" module for general maintainability
- Defines RootState TypeScript interface composed of Vuex store modules' states (required after moving consents module to separate file)


JuroOravec:
Good points. Originally I've put the typescript interface refactoring as it was required, but it's true that although that issue was blocking this one, they are different. That PR is here https://github.com/DD-DeCaF/caffeine-vue/pull/228. Could you give it a review? This PR is blocked by that one.

ChristianLieven:
Nice that you cleaned this up! Merge away

JuroOravec:
Done! And kudos to Ali for helping me with the commits 🎉

PR #228, opened 2020-02-04
refactor: define RootState TypeScript interface

JuroOravec:
The interface is necessary for using rootState inside store modules.
Otherwise the compiler complains.


ChristianLieven:
Looking good! Merge away!

PR #229, opened 2020-02-05
Fix/cookies policy too strict

JuroOravec:
Fixes issue when user is logged out in the case that:
1) User is initially logged out.
2) Before logging in they changed their cookie consents to reject preferences (e.g. by not ticking the preferences option in the cookie banner).
3) They log in.
4) They changed their preferences cookies to accepted as logged in.

This action should keep the user logged in, as they agreed to have their login token remembered.

The issue was with vanilla JS events. Event callback to clear localStorage on "beforeunload" was assigned multiple times, but removed only once, which meant that only one of callbacks was removed, but others still cleared the localStorage despite user's state permitting the storage.

Additionally made the warning stating the implications of rejecting preferences cookies more prominent.

### DIFF
--- a/src/components/CookieConsent.vue
+++ b/src/components/CookieConsent.vue
@@ -37,8 +37,9 @@
             <router-link to="privacy-policy">Privacy policy</router-link>.
           </p>
           <p class="mb-0">
-            <small>
-              Note: Preferences cookies are necessary to hide this notification.
+            <small class="warning py-1 px-1 font-weight-bold">
+              Note: Preferences cookies are necessary to hide this notification
+              and remember login.
             </small>
           </p>
         </v-container>
--- a/src/store/modules/consents.ts
+++ b/src/store/modules/consents.ts
@@ -167,8 +167,9 @@ export default vuexStoreModule({
         message:
           "These cookies allow to remember choices you have made such as " +
           "what your user name and password are so you can automatically " +
-          "log in.<br/><br/>Note: Preferences cookies are necessary to " +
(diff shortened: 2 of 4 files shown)


ChristianLieven (approved):

ChristianLieven:
Good to go imo! Feel free to merge!

kvikshaug (commented):
Hmm, why do we need a vuex plugin to add/remove the event listener? Couldn't it be defined in `App.vue` or something, always run, and check the store stage at the time it runs? I don't see the difference between the sanity check, and the conditionals that decide whether to add/remove the event listener.
(Sorry, I realize that wasn't introduced in this PR, but it's the first time I'm seeing it now.)

JuroOravec:
@kvikshaug Good suggestion, I like it!

Originally I've done it as a Vuex plugin to listen to preferences cookie consent changes, and the "gatekeeper" logic that decided whether the `clearPreferencesCookies` function would run was whether the event listener exists or not.

But it's true that since that was unreliable (multiple events attached), and now the decision falls upon the result of `store.getters["consents/isConsentAccepted"]`, the adding/removing of events can be removed, we don't need to constantly listen for changes, and so it doesn't have to be a vuex plugin.

kvikshaug (approved):
As for the changes in this PR, looks good!

JuroOravec:
Merged this one. And the PR for moving the logic from Vuex plugin is here https://github.com/DD-DeCaF/caffeine-vue/pull/232

PR #230, opened 2020-02-05
Feedback simulation

blablabananarama:
Tabbed interface on Data-Driven cards created, which shows errors/ warnings.


ChristianLieven (approved):
Looks mighty sexy! 

One last thing I wanna suggest is to give Warnings and Errors in the Tab title a badge each!

Otherwise, I approve!

blablabananarama:
The first badge renders over the second one on the simple implementation that I have come up with, will leave as is and merge.

PR #232, opened 2020-02-05
refactor: move cookie clearing from vuex to vue

JuroOravec:
PR for the [suggestion to move localStorage clearing logic from Vuex plugin](https://github.com/DD-DeCaF/caffeine-vue/pull/229#pullrequestreview-353781253) from https://github.com/DD-DeCaF/caffeine-vue/pull/229.


kvikshaug (approved):
Thanks for looking into this! :+1:

PR #236, opened 2020-02-10
Feat/analytics

JuroOravec:
- Introduces analytics store module.
- Introduces analytics Vue plugin with auto page tracking and directive `analytics-model` to trigger analytics store module actions on `v-model` change.
- Replaces VueAnalytics with [analytics](https://github.com/DavidWells/analytics). GA tracking implemented via [analytics ga plugin](https://getanalytics.io/plugins/google-analytics).
- Introduces `a-extended` component for link tracking.
- Tracking is toggable via store and via env vars (same as consents, off by default, and on on staging and prod).
- Tracking is disabled on `DoNotTrack` browser setting, when analytics are disabled, or when `"statistics"` cookie consent is not given.
- Introduces bunch of analytics plugins for enriching/modifying the payload, and bunch of helper functions to add/remove these plugins, in `src/utils/analytics`
- Introduces packages `analytics`, `analytics-plugin-do-not-track`, `analytics-plugin-ga`, `analytics-plugin-original-source`
- Tracks following events:
  - `search_escher` - when user searches escher (and the keyword doesn't change for 2 sec)
  - `send_to_visualize` - when predictions are sent for visualization
  - `visualize` - when Escher map is being (re)loaded. Describes which card/map/project/etc is being visualized.
  - `export` - when user exports item(s)
  - `link` - when user clicks on an external link (external link is defined as having different hostname)
  - `track_dialog` - when a new dialog is open (new model, new project, new map, card, ...)
  - `page` - when user navigates to different page
  - `login` - when user logs in (either via social/email login or from localStorage)
-  on registration or login (either via social/email login or from localStorage), user is identified by the following (depending on availability): internal user id, firebase id, or hashed email
- on registration or login (either via social/email login or from localStorage), user info is captured (depending on availability)
- on logout, user is forgotten and is treated as an anonymous user.


src/utils/utility.ts
```diff
@@ -104,3 +104,11 @@ export function camelCasePropertyNames(obj) {
...
   return moment(datetime).format("YYYY-MM-DDTHH:mm:ss.SSSSSZ");
 }
+
+export function tryCatchCall(fn, thisArg, ...args) {
+  try {
+    return fn.call(thisArg, ...args);
+  } catch (e) {
+    console.warn(e);
+  }
+}
```
kvikshaug: This is an odd function, I recommend silencing explicitly where needed.

kvikshaug (changes requested):
- The analytics package [seems to be developed by a single contributor](https://github.com/DavidWells/analytics/graphs/contributors), so the project has a high bus-factor. I'm also not seeing a huge community around this dependency so I recommend to reconsider using it.
- The plugins are a huge introduction of low-level code and I am worried about maintainability. Isn't it possible to bring these in as dependencies without adding the code in our repo?
- I am not sure we can introduce MIT licensed code in our repository as all our code needs to be open sourced under the Apache license.
- I'm a bit on the fence about the `a-extended` tag, I think it's a good abstraction but it can be hard for new developers to realize they need to use it over a regular anchor in order to track a click. Maybe documentation would solve that.

As for the Travis build, you need to run the linter, and it also looks like there are a number of typescript issues that need to be resolved.

JuroOravec:
> * The analytics package [seems to be developed by a single contributor](https://github.com/DavidWells/analytics/graphs/contributors), so the project has a high bus-factor. I'm also not seeing a huge community around this dependency so I recommend to reconsider using it.

It's true about the single contributor, but the use also seems to be growing steadily since Oct 2019 where it seems to have got some recognition or was put into use. The interaction with that package is limited to the inside of the analytics store module, and the package itself depends only on redux, so I think it's fairly safe. So I suggest to use it, and if push comes to shove and it really gets broken and forgotten, then replacing the calls to that package with direct calls to e.g. Google Analytics should be fairly simple.

> * The plugins are a huge introduction of low-level code and I am worried about maintainability. Isn't it possible to bring these in as dependencies without adding the code in our repo?
> * I am not sure we can introduce MIT licensed code in our repository as all our code needs to be open sourced under the Apache license.

Hm, how exactly does licensing work? I thought there's no difference if you use a package as a dependency to your source code vs when you have the actual code in your code base? 

So do I understand correctly that you're saying that if the MIT-licensed code lived on its own, its distribution could be used as a dependency without issues?

Also, about the maintainability, are you pointing at the code that manages/creates the plugins, or with the plugins included?  I'm asking because half of the custom plugins are specific to our use case (`requireConsentPlugin` and `disableAnalyticsPlugin`, `enrichAnalyticsPlugin`), so in their case it'd make more sense to do those actions on the payload inside the store before it's passed to the analytics than have them as dependencies. For others, like `dropNoValuePropertiesPlugin` `snakecasePropertiesPlugin`, these are general enough that they could be separate.

So if I understand correctly, it would be desirable to extract the code that encapsulates `VueAnalytics` and `analytics` packages, along with other related helper functions, and have that as a separate dependency? And then also package separately `dropNoValuePropertiesPlugin` and `snakecasePropertiesPlugin`?

> * I'm a bit on the fence about the `a-extended` tag, I think it's a good abstraction but it can be hard for new developers to realize they need to use it over a regular anchor in order to track a click. Maybe documentation would solve that.

In the spirit of documenting up how things work, it might be actually good to have a doc folder that would explain how the frontend app is wired up, and there it could also include custom components like the `a-extended`? What do you think?

> As for the Travis build, you need to run the linter, and it also looks like there are a number of typescript issues that need to be resolved.

Yes, that was likely also the cause of the issue I was getting. Fixed the type issues and all is fine now. But it's weird that Travis reported missing dependencies as the cause of failure instead of the actual type errors.

src/utils/utility.ts
JuroOravec: I removed it. Originally, it was meant to silence GA analytics plugin error on invalid trackingID, but it turned out that the plugins are initialized asynchronously, so no need for the try/catching.

JuroOravec:
I trimmed and rewrote the logic and replaced licensed files. The analytics package plugins, vue directive and vue plugin are now all in `src/plugins/analytics.ts`

.env.staging
```diff
@@ -2,4 +2,5 @@ VUE_APP_ENVIRONMENT=staging
 VUE_APP_SENTRY_DSN=https://44d530ac9cb14ad391a54689b2286a92@sentry.io/1460040
 VUE_APP_DEPLOYMENT_HASH=$TRAVIS_COMMIT
 VUE_APP_GA_TRACKING_ID=UA-106144097-3
-VUE_APP_ENABLE_CONSENTS=true
+VUE_APP_ENABLE_ANALYTICS=true
+VUE_APP_ENABLE_CONSENTS=true
```
kvikshaug: Files should end with a newline

src/utils/utility.ts
```diff
@@ -1,4 +1,4 @@
-import Vue from "vue";
+import * as Sentry from "@sentry/browser";
```
kvikshaug: Looks like an unused import

kvikshaug (approved):
Thanks for the updates!

.env.staging
JuroOravec: fixed

src/utils/utility.ts
JuroOravec: fixed. Thanks for spotting it. I don't actually remember why it was imported in the first place.

PR #239, opened 2020-02-11
Media crud - final touches

kvikshaug:
@ChristianLieven I created a separate branch so that you can review only my changes - could you look over this? When done I'll merge both branches.

I have not looked into the identifier issues yet, so created media will not work well with our models in the current state. I think we need to do some refinement/brainstorming on that issue. Issue on that [here](https://app.zenhub.com/workspaces/bioviz-scrum-5d22fe4139b1324e0011234c/issues/dd-decaf/scrum/1280), please prioritize as you see fit - but I'd recommend we discuss it this sprint if community modeling with custom medium is going to be part of the review.

### DIFF
--- a/src/store/index.ts
+++ b/src/store/index.ts
@@ -80,7 +80,7 @@ export default new Vuex.Store({
       dispatch("projects/fetchProjects");
       dispatch("strains/fetchStrains");
       dispatch("media/fetchMedia");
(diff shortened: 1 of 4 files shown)


ChristianLieven:
Nicely done! I knew I was tripping up somewhere, I'm glad you spotted that it was on object reassignment.

PR #244, opened 2020-02-14
Explain steadycom

ChristianLieven:


ChristianLieven:
> Looks good, nothing to complain, my only thought was is the little bit of text under the selector still necessary?:
> grafik

I think it's nice for a quick summary! But I've adapted the text a bit in my latest commit!

PR #245, opened 2020-02-14
feat: store BiGG identifiers for medium compounds

kvikshaug:


kvikshaug:
> LGTM, tho can the `$event.annotation['bigg.metabolite']` ever be `undefined` ?

Yes, definitely it can for metabolites that don't have xrefs to bigg. I'll put up a guard.

PR #246, opened 2020-02-15
Review fixes

ChristianLieven:
Quick fixes:
- removal of mentioning CPLEX
- Adding edge highlights to the cross feeding graph

### DIFF
--- a/src/views/CommunityModeling.vue
+++ b/src/views/CommunityModeling.vue
@@ -483,7 +483,9 @@ export default Vue.extend({
 
       // link template
       var linkTemplate = chart.links.template;
-      linkTemplate.strokeOpacity = 0;
+      linkTemplate.stroke = am4core.color("#000");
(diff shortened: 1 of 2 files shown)


kvikshaug (commented):
Codewise it looks good but I tried rendering locally and it looks like this, it doesn't quite seem that this is what we intended wrt the edges?

image

kvikshaug (approved):
Ah, I see there is a slight edge now but it's not so easy to see. I'll approve in terms of the code, but myabe consider asking for design feedback to see what others think?

ChristianLieven:
> Ah, I see there is a slight edge now but it's not so easy to see. I'll approve in terms of the code, but myabe consider asking for design feedback to see what others think?

I've played a little with the opacity. It's a bit hit and miss as overdoing it quickly gets a pure black block for the smaller transactions.

PR #250, opened 2020-02-24
Feat/hotjar

JuroOravec:
Add Hotjar enabling/disabling logic and Hotjar init script.


src/utils/settings.ts
```diff
@@ -17,9 +17,16 @@ export const apis = {
 export const trustedURLs = process.env.VUE_APP_TRUSTED_URLS.split(",");
 export const sentryDSN = process.env.sentryDSN;
 export const gaTrackingID = process.env.VUE_APP_GA_TRACKING_ID;
+export const hotjarID = Number.parseInt(process.env.VUE_APP_HOTJAR_ID);
```
kvikshaug, line 20: Won't this error if the id isn't specified, as I assume it won't be in dev mode?

kvikshaug (approved):
Looks good!

src/utils/settings.ts
JuroOravec: That specific line won't. It'll be just `NaN`.  Where this would throw error down the line if the ID would not be given is when the HotJar init script will attempt to fetch the actual HotJar script and the invalid URL will return 404 (see `src/store/modules/analytics.ts.actions.loadHotjar`). But HotJar is disabled in dev, so the init script will not run at all.

PR #251, opened 2020-02-24
refactor: remove all cookies

JuroOravec:
With analytics, more cookies have been added, so it make more sense to
remove all cookies at once if they cannot be stored on user's machine than doing so selectively.

### DIFF
--- a/src/store/modules/consents.ts
+++ b/src/store/modules/consents.ts
@@ -245,9 +245,7 @@ export default vuexStoreModule({
           category: "preferences"
         });
         if (!consentGiven) {
-          ["jwt", "cookie:accepted", "consents"].forEach(k =>
-            localStorage.removeItem(k)
-          );
+          localStorage.clear();
         }
       });
     },


kvikshaug (commented):
I don't agree there is and always will be a 1-1 relationship between items in localstorage and user consents. I think for example some LS key is used by webpack for development which may be affected by this.

JuroOravec:
Hm, good point. Though on the other hand, consents are disabled in dev by default, so it wouldn't be affected there. But I agree that it could create issues in the future. Plus being explicit about the values removed from localStorage can serve as a sort of documentation.

src/store/modules/consents.ts
```diff
@@ -245,9 +245,7 @@ export default vuexStoreModule({
           category: "preferences"
         });
         if (!consentGiven) {
-          ["jwt", "cookie:accepted", "consents"].forEach(k =>
-            localStorage.removeItem(k)
-          );
+          localStorage.clear();
```
ChristianLieven, line 248: So with Ali's comment and your response in mind, do you plan to change this to explicitly remove items?

ChristianLieven (approved):

src/store/modules/consents.ts
JuroOravec: Yup, I'll close this one and create a new PR which will have all the items for deletion specified.

ChristianLieven: OK, go ahead!

PR #253, opened 2020-02-25
fix: error on updateUser action

JuroOravec:
Calling analytics.identify with `null` evaluated `null` as valid ID
(error in analytics package). Changing `null` to "" evaluates it
correctly as no ID.
Additionally chained updateUser calls after identifyUser, so the user ID
is always set.


kvikshaug (approved):
Nice catch! :rocket:

PR #254, opened 2020-02-25
chore: update cookie removal

JuroOravec:
Updates info that should be cleared from user's browser and expanded it with sessionStorage and document.cookie

### DIFF
--- a/src/store/modules/consents.ts
+++ b/src/store/modules/consents.ts
@@ -245,9 +245,49 @@ export default vuexStoreModule({
           category: "preferences"
         });
         if (!consentGiven) {
-          ["jwt", "cookie:accepted", "consents"].forEach(k =>
-            localStorage.removeItem(k)
-          );
+          // Clear localStorage
+          if (window.localStorage) {
+            [
+              // Session Token
+              "jwt",
+              // Vuex consents store module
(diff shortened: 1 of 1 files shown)


src/store/modules/consents.ts
```diff
@@ -245,9 +245,49 @@ export default vuexStoreModule({
...
+              // Google Analytics
+              "_ga",
+              "_gid",
+              // HotJar
+              "_hjid",
+              "_hjIncludedInSample"
+            ].forEach(
+              k =>
+                (document.cookie =
+                  k + "=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;")
```
kvikshaug, line 288: Is this really how to expire a cookie from javascript? :eyes:

kvikshaug (approved):
Thanks for adding the comments, they make it very readable :+1:

src/store/modules/consents.ts
JuroOravec: Yup, you set an expiry time to the past. [Also suggested by MDN ](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#Example_4_Reset_the_previous_cookie)

